PR #12, opened 2023-02-28
add computed files to gitignore

sheridansunier:
chore(added computed files to gitignore): add computed files to gitignore


jedjedjedM:
@sheridansunier : Given the clarity from @sanrai on the `dist` folder, I think this should probably be closed.

PR #47, opened 2023-03-29
MWPW-128673: fix: brings page back to top of collection on page change

cmiqueo:
Description:
When changing pages while using the paginator the page scrolls to the top of card collection container:

Resolves:
https://jira.corp.adobe.com/browse/MWPW-128673

Video Demo:
https://jira.corp.adobe.com/secure/attachment/9859441/ScrollToTop.mp4


sanrai (changes requested):
Should this feature be disabled/updated for mobile?

Since mobile lays out the cards one at a time, so if you have a paginator and go down to card 25 and then click next page, you get scrolled all the way to the top and need to scroll back down (which i think is the UX we are trying to prevent)

See video:
https://www.vidline.com/share/V0EVHNVW46/fc2bdedd1bc7c800811f82ed85aa1001

PR #56, opened 2023-04-06
fix(Paginator.jsx): prevents paginator to scroll if not needed [MWPW-129225]

cmiqueo:
Summary: 
Check the position of the card collection container, and scrolls the age only if needed

Resolves: 
https://jira.corp.adobe.com/browse/MWPW-129225

Move demos
Before:
https://jira.corp.adobe.com/secure/attachment/9907983/scroll%20-%20before.mp4

After:
https://jira.corp.adobe.com/secure/attachment/9907984/scroll%20-%20after.mp4


sanrai:
Hi @cmiqueo,

You are going to need to show me how to test this. 

I pulled the changes locally and don't see any difference in functionality with or without the code snippet you added for the behavior with respect to the videos shared.

cmiqueo:
@sanrai 
I think the videos demo the before and after. But basically to test:

Go to the Milo configurator
Advanced > Reset to default state
Basic > Results Per Page: 2 or 3
Pagination > Enable Pagination > Paginator
Click on any page in the paginator.

Assuming you were at the top of the page, you will noticed how the page scrolled when is not needed since there should be only one row of cards and it is fully visible

sanrai:
Hi @cmiqueo,

I've tried the steps you've given above as well as looked at your demo videos and I am not able to recreate this issue locally and from my side don't see any functionality differences with and without your PR.

PR #58, opened 2023-04-13
fix: resets current page state when searching (MWPW-129459)

cmiqueo:
1. Resets the `state.currentPage` value when performing searches 
2. Adds new test page for paginator

Motivation
Prevent errors when the users search on a page number larger than the actual results.
 
Resolves: https://jira.corp.adobe.com/browse/MWPW-129459

Test page
https://adobecom.github.io/caas/html/paginator.html

Test steps:
1 Navidate to any page hihger that
2 Enter **red** in the _Search_ input field

Notice the paginator state


sanrai:
PR looks good to me and passes all testing. 

Only two pieces of feedback:

1.  Rename title to `fix(Container.jsx): resets current page state when searching [MWPW-129459]`
2.  I would recommend renaming `paginator.html` to `MWPW-129459.html` so it becomes easier to find regress this issue in the future using Jira.

PR #59, opened 2023-04-14
fix(Container.jsx): resets current page state when searching [MWPW-129459]

cmiqueo:
1. Resets the state.currentPage value when performing searches
2. Adds new test page

Motivation
Prevent errors when the users search on a page number larger than the actual results.

Resolves: https://jira.corp.adobe.com/browse/MWPW-129459

Test page
https://adobecom.github.io/caas/html/carlos-test-page.html

Test steps:
1 Navigate to any page higher than 5
2 Enter **red** in the _Search_ input field

Notice the paginator state


sanrai:
Hi @cmiqueo, 

1. The e2e PR check is failing. I think updating chromedriver to `111.0.0` from `110.0.0` may fix iti

image

2. Please update PR title to `fix(): resets current page state when searching`

PR #72, opened 2023-06-01
test(MWPW-130075): Card with hidden CTA should be clickable

gmirijan:
1. Add automation test: Card with hidden CTA should be clickable
2. Resolve config clone issue by replacing shallow clone with deep clone 

- Add test tickets to the Hide CTA(s) test suit
- Update chromedriver
Resolves test ticket (second part):
https://jira.corp.adobe.com/browse/MWPW-130075

Story tickets:
https://jira.corp.adobe.com/browse/MWPW-126169 - Hide Individual CTA


e2e-tests/specs/all.e2e.js
```diff
@@ -274,23 +272,35 @@ describe('Hide CTA(s):', async () => {
...
+    });
     it('MWPW-128711: ALL CTAs should be hidden', async () => {
-        const cloneConfig = { ...config };
+        const cloneConfig = structuredClone(config);
         cloneConfig.collection.collectionButtonStyle = 'hidden';
         const state = btoa(JSON.stringify(cloneConfig));
         const url = `${serverPath}/html/e2e/MWPW-126169.html?state=${state}`;
         await browser.url(url);
         await browser.setTimeout({ script: 50000 });
+        await browser.pause(5000);
```
sanrai, line 292: @gmirijan is this browser pause of 5 seconds necessary for the e2e test or just a left over while developing? Best would be if we could remove it to remove it since it'll add 5 seconds extra to our build process.

sanrai (approved):

e2e-tests/specs/all.e2e.js
gmirijan: It is a left over and should be removed.

sanrai (approved):

PR #75, opened 2023-06-28
fix: changes card backgroundImage to IMG tag [MWPW-129969]

cmiqueo:
Description:
This PR fixes the problem that some cards were missing the background image if there was duplicated images and dates: 
https://business.adobe.com/customer-success-stories.html?page=7
https://business.adobe.com/resources/main.html

Moves the main card image from backgroundImage css property to a IMG tag with "loading=lazy" attribute and removing the need for the complexity of lazy loading using intersection observer.

Resolves: 
https://jira.corp.adobe.com/browse/MWPW-129969


sanrai (approved):
This is a great idea/solution and much better implementation than what previously existed 👍

PR #94, opened 2023-08-30
feat(MWPW-135383): deep linking filters && clear all for search url 

sheridansunier:
In https://jira.corp.adobe.com/browse/MWPW-119624, updates were made to include support of deep linking and clicking back button when a search term is used in a CaaS card collection.  

We discovered that filter deep linking (e.g. https://business.adobe.com/resources/main.html?ch_Products=Marketo%2520Engage) doesn't work.

We also discovered that the "clear all" button did not remove search terms from the url.

This PR contains the fixes for both issues


sanrai:
@sheridansunier , could you squash all your commits into one and message it based on the following description. 

image

Also update the title of the PR to match.

Otherwise our release process won't version this correctly.

sanrai (changes requested):
@sheridansunier , could you squash all your commits into one and message it based on the following description. 

image

Also update the title of the PR to match.

Otherwise our release process won't version this correctly.

Source: https://github.com/conventional-changelog/commitlint

PR #100, opened 2023-09-19
feat(MWPW-125698): lana logging

sheridansunier:
Integrates CaaS with Lana and adds a first pass of lana logs.

The primary focus of these logs is to add logging around our collection endpoint calls.  I also added some logging around try/catch blocks within reason.  After we have time seeing the implicit logging we get we can choose what kind of coverage and logs we want to add.


react/src/js/components/.DS_Store
cmiqueo, line 1: WE should `.DS_Store` to the .gitignore list

cmiqueo (approved):

react/src/js/components/.DS_Store
sheridansunier: ok silly question... but once I have updated .gitignore is there a good way to remove those .DS_Store files from this PR?

react/src/js/components/Consonant/Helpers/lana.js
```diff
@@ -0,0 +1,34 @@
...
+            window.lana.log(e.reason || e.error || e.message, { errorType: 'i' });
+        }
+    };
+
+    window.lana = {
+        log: async (...args) => {
+            window.removeEventListener('error', lanaError);
+            window.removeEventListener('unhandledrejection', lanaError);
+            // eslint-disable-next-line import/no-unresolved, import/extensions
+            await fetch('https://milo.adobe.com/libs/utils/lana.js');
```
jedjedjedM, line 15: This should load from `www.adobe.com/libs/utils/lana.js` instead of milo.adobe.com

sheridansunier: ok updated

PR #101, opened 2023-09-20
fix(mwpw-136323): when you clear a filter group it now updates the url

sheridansunier:
Filter param was not removed from URL when the filter cleared by clicking on the number next to the filter group, nor was it removed when you clicked the checkbox to clear a filter.  This PR addresses this bug.


react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -429,17 +429,19 @@ const Container = (props) => {
...
         });
 
         const urlParams = new URLSearchParams(window.location.search);
         clearUrlState();
+        // actually clear the url state
         urlParams.forEach((value, key) => {
             const chFilter = key.toLowerCase().replace('ch_', '').replace(' ', '-');
-            if (key.indexOf(filterGroupPrefix) !== 0 || !id.includes(chFilter)) {
+            if (key.indexOf(filterGroupPrefix) !== 0 || !id.toLowerCase().includes(chFilter) ||
+                !group.toLowerCase().includes(chFilter)) {
```
cmiqueo, line 444: Since there are more than two conditions here I would go with the following for better readability
```
if (key.indexOf(filterGroupPrefix) !== 0 
    || !id.toLowerCase().includes(chFilter)
    || !group.toLowerCase().includes(chFilter)) {
```

cmiqueo (approved):
Approved with small comment :)

PR #103, opened 2023-09-27
test(multiple): added e2e tests for filter and search behaviors

sheridansunier:
https://jira.corp.adobe.com/browse/MWPW-137132
https://jira.corp.adobe.com/browse/MWPW-136771
https://jira.corp.adobe.com/browse/MWPW-137140
https://jira.corp.adobe.com/browse/MWPW-137141
https://jira.corp.adobe.com/browse/MWPW-137142

PR #104, opened 2023-10-12
fix(mwpw-136333): filter params were not removed with number and clea…

sheridansunier:
…r all was not clearing search

actually finishes the work for [MWPW-13623](https://jira.corp.adobe.com/browse/MWPW-136323) and [MWPW-136333](https://jira.corp.adobe.com/browse/MWPW-136333)

should be merged after https://github.com/adobecom/caas/pull/103 so I can update the e2e tests to catch these and so the entire suite can run on this as a test.


e2e-tests/specs/all.e2e.js
```diff
@@ -368,8 +368,8 @@ describe('Filter and Search Behaviors', async () => {
         await clearAll.click();
         const cardRes = await $('.consonant-LoadMore-text').getText();
         expect(cardRes).toEqual('Showing 5 of 8 cards');
-        // const newUrl = await browser.getUrl();
-        // expect(newUrl).toEqual(`${serverPath}/html/e2e/e2e-grid.html?state=${state}`);
+        const newUrl = await browser.getUrl();
+        expect(newUrl.replaceAll('%3D', '=')).toEqual(`${serverPath}/html/e2e/e2e-grid.html?state=${state}`);
```
cmiqueo, line 372: Instead of using a replace() function, why not just change the expected string:
`expect(newUrl).toEqual(`${serverPath}/html/e2e/e2e-grid.html?state%3D${state}`);`

react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -476,7 +476,8 @@ const Container = (props) => {
         const urlParams = new URLSearchParams(window.location.search);
         clearUrlState();
         urlParams.forEach((value, key) => {
-            if (key.indexOf(filterGroupPrefix) === -1) setUrlState(key, value);
+            if (key.indexOf(filterGroupPrefix) === -1 &&
+                key.indexOf(searchPrefix) === -1) setUrlState(key, value);
```
cmiqueo, line 480: How about this formatting (for consistency):
```
 if (key.indexOf(filterGroupPrefix) === -1 
    && key.indexOf(searchPrefix) === -1) setUrlState(key, value);
```

cmiqueo (approved):
Approved with some minor suggestions :)

e2e-tests/specs/all.e2e.js
sheridansunier: because the equals signs I'm replacing that are causing me issues are two ones that get postpended into the state when it is parsed specifically by the github production version of the e2e tests - there are no computed %3Ds in the local tests so changing the expected string breaks local tests

PR #117, opened 2023-11-21
fix(mwpw-138742): filter-params removed from url when group name has a space

sheridansunier:


react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -443,7 +443,7 @@ const Container = (props) => {
             const chFilter = key.toLowerCase().replace('ch_', '').replace(' ', '-');
             if (key.indexOf(filterGroupPrefix) !== 0
                 && !id.toLowerCase().includes(chFilter)
-                || !group.toLowerCase().includes(chFilter)) {
+                || !group.toLowerCase().replace(' ', '-').includes(chFilter)) {
```
cmiqueo, line 446: This will replace only the first space, you should use replaceAll(),
Try it in the console
`'a b c d'.replace(' ', '-')` vs. `'a b c d'.replaceAll(' ', '-')`

cmiqueo (changes requested):

react/src/js/components/Consonant/Container/Container.jsx
sheridansunier: good catch

gmirijan (approved):

cmiqueo (approved):

PR #128, opened 2024-03-04
DOTCOM-114584: Fixes a badge swap timing bug

dol-iva:
**Bug Description:**
There is a minor bug happening when transitioning from the end of an event to the next variation when using servertime and DXFs.

**Short Explanation:** When using servertime and DXFs, the variation transition happens at the same time as the end of the event. Since the DXF transition, and the event ending happen at the same time, in many cases the ending event keeps the "live" badge instead of changing to the "on demand" badge.

NOTE1: This is my first time working with this codebase, so let me know if something doesn't look good.
NOTE2: There might be another way of fixing the issue by giving the `differential` variable a jumpstart value (eg. 1000 instead of 0) 

**Before Fix**
image
image

**After Fix**
image


dol-iva:
Having issues fixing the merge conflicts. I'll create a new PR.

PR #156, opened 2024-06-03
feat(MWPW-151628): Adds support for new categories layout

cmiqueo:
Adds support for the categories layout used by the T3 MVP Events page

Resolves:
https://jira.corp.adobe.com/browse/MWPW-151628


react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -244,6 +250,8 @@ const Container = (props) => {
      * @type {[Array, Function]} Filters
      */
     const [filters, setFilters] = useState([]);
+    window.filters = filters;
```
sanrai, line 253: Should `window.filters` not be exposed?

sanrai (approved):

PR #163, opened 2024-07-22
MWPW-153580: Update Badge Display Logic Based on Content Origin

sanrai:
## Description
This pull request addresses an issue where Milo was incorrectly sending the `primaryTag` information for each card. 

Specifically, for Milo (using the postXDM service), the badge label information was being sent incorrectly as an ID, missing the tag's title and localized values.

To resolve this, we will use the tag data inside Chimera IO to look up the correct titles based on the ID we get from Milo. 

This approach is easier than fixing the postXDM service and migrating the database in OpenSearch.

This approach, however, makes it so we have to handle badges differently based on the content origin. 

We could have collections from a Dexter origin where these need to have badge labels to show up (since they explicitly authored primaryTags for badges to show up).

While for Milo, authors may or may not expect badging to show up. 

By default, they will not want it, since they could've authored a collection and accidentally put a primaryTag field without realizing that would cause badges to show up.

On the other hand, other Milo authors could put the primaryTag and want badges to show up.

So we need to give them an option to opt-in, and once they do, badges will show up.

The logic: Dexter content should always display badges by default, while Milo content should only display badges if the author opts in. 

This prevents unintended badge displays for Milo authors who did not expect them.

## Solution

1. Handling Different Content Origins:

- For content from Dexter, badges should always show (default behavior).
- For content from Milo, badges should only show if the author has opted in. 

2. Code Changes:

- Introduced a check for the content origin.
- If the content origin is Dexter, badges are shown by default.
- If the content origin is Milo, badges are shown only if the author has opted in.

### Updated Code
```
const showCardBadges = getConfig('collection', 'showCardBadges');
const fromDexter = origin === 'Dexter';
const showBadge = (isOneHalf || isThreeFourths || isFull) && (fromDexter || showCardBadges);
```

### New Logic Implementation
```
if (origin === 'dexter' || optedIn) {
    showBadge;
}
```

### Additional Notes
- Ensure that the list of origins is maintained to check if the content is from Dexter or Milo _(on Chimera IO side)_
- This change encapsulates the badge display logic, ensuring that the correct tags are shown based on the content origin and author preference.

### Testing
- Verify that badges are displayed correctly for Dexter content by default.
- Verify that badges are displayed for Milo content only when opted in by the author.
- Ensure that no badges are shown for Milo content if the author has not opted in.

By merging this pull request, we will correct the badge display issue, ensuring that authors have control over badge visibility based on the content origin and their preferences.


gmirijan (approved):

gmirijan (approved):

cmiqueo (approved):

sheridansunier (approved):
verified this w the planned changes on caas

PR #170, opened 2024-08-13
MWPW-145727: add ability to have secondary cta used during live events

sheridansunier:
As a Virtual Events Product Manager, I want to have a way for a different destination (e.g anchor link) and CTA copy when an event card is live so that when we use event cards to promote the upcoming homepage schedule, a click to the card will take the user to the specific destination (specifically to the marquee video player on the homepage).

This change requires: 
- https://github.com/adobecom/milo/pull/2707
- https://git.corp.adobe.com/wcms/chimera-io/pull/435

To test this, you need a card that has a a secondaryCTA authored, and you need it to be an event that you can use servertime to observe the cards behavior before/during/after the card goes live.  For my personal testing I was using this card collection: [here it is!](https://14257-chimera-feature.adobeioruntime.net/api/v1/web/chimera-0.0.1/collection?originSelection=northstar&contentTypeTags=caas%3Acontent-type%2Fevent-session&secondSource=&secondaryTags=&collectionTags=&excludeContentWithTags=caas%3Aevents%2Fmax%2Ccaas%3Acontent-type%2Fproduct&language=en&country=US&complexQuery=&excludeIds=64c7ad2e-7bf1-5e7f-8619-c66f27371445%2C0dfd22b4-56d1-5ce5-ad6e-b58c5ef608d5%2Cd69bc403-58d5-5fcf-9fad-4d006f6b96d9&currentEntityId=&featuredCards=&environment=&draft=false&size=4500&flatFile=false ) . But this does not show the actual feature working e2e b/c I mocked the secondaryCTA in my personal development. 

You need to check that altCTA is used from milo.  Be using the chimera-feature (or dev if it has merged but if it hasn't the most up to date code is on chimera-feature) and then you should be able to see that the CTA changes to the secondaryCTA when a card is live.


sanrai (approved):
LGTM! :) Well done. 

I was trying to simplify `right && right.length > 0` but couldn't find a cleaner way to do it than you have.

PR #173, opened 2024-09-09
feat(mwpw-157435): alternative cta links/text for live & upcoming

sheridansunier:
[MWPW-145727](https://jira.corp.adobe.com/browse/MWPW-145727) was recently released and in preparation of this feature, we started the configuration of the Card Collection that would use this feature on the MAX Miami Homepage:

We discovered that the default CTA for caas:content-type/event-session is "Watch now". In the original [MWPW-145727](https://jira.corp.adobe.com/browse/MWPW-145727) I had assumed that CTA was "Learn more". 

We would like to ensure the following CTA copy & destinations based on state:
 Upcoming

State | CTA Copy | Destination | Destination Example
-- | -- | -- | --
Upcoming | Learn More | Card URL | https://www.adobe.com/max/2024/sessions/creativity-super-session-motion-design-for-graphic-ss2.html
Live | Watch Now | IF LIVE CTA functionality is enabled Marquee | #
Live | Watch Now | Card URL | https://www.adobe.com/max/2024/sessions/creativity-super-session-motion-design-for-graphic-ss2.html
On Demand | Watch Now | Card URL | https://www.adobe.com/max/2024/sessions/creativity-super-session-motion-design-for-graphic-ss2.html

Because "Watch Now" is the default state of the Event Session Card, we should only update the CTA URL when a Session is Live (using Secondary CTA URL) and update the CTA label / copy when a Session is Upcoming (using Secondary CTA Label). This implies we can remove the change to CTA label when the card is live from the original feature 

Just documenting this because right now when a card is Live it updates the CTA Copy & Label, so we'll want to tweak that


cmiqueo (approved):
I would use `altRightLive `and `altRightUpcoming `
`altRightA `and `altRightB`

sheridansunier:
@cmiqueo good note - changed the names :)

PR #186, opened 2024-10-04
feat(MWPW-150610): implement partial load with background fetch

sanrai:
# Performance Optimization for Large Card Collections
## Problem
Large card collections (up to 4000 cards) were causing slow initial page load times and poor user experience.
## Solution
This PR implements a two-phase loading strategy:
1. Initially load a smaller subset of cards (e.g., 100) for immediate display.
2. Load the remaining cards in the background after the initial render.
## Changes
- Added new configuration options in `index.html`:
  - `partialLoadWithBackgroundFetch.enabled`
  - `partialLoadWithBackgroundFetch.partialLoadCount`

- Updated `Container.jsx` to implement the new loading strategy:
  - Added state management for total card count
  - Modified `getCards` function to support partial loading
  - Updated pagination and load more functionality to work with total card count

- Adjusted various component props to use `cardCount` instead of `gridCards.length`
## How to Test
1. Configure a large card collection (1000+ cards)
2. Enable `partialLoadWithBackgroundFetch` in the configuration
3. Set `partialLoadCount` to a smaller number (e.g., 100)
4. Load the page and observe:
   - Initial quick load with subset of cards
   - Background loading of remaining cards
   - Correct total card count in pagination/load more
## Notes
- This feature is optional and can be enabled/disabled via configuration
- The `partialLoadCount` is configurable to allow fine-tuning based on specific use cases


react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -1350,7 +1358,7 @@ const Container = (props) => {
                 <section
                     ref={box}
                     role="group"
-                    aria-label={title || 'Card Collection'}
+                    aria-label={title}
```
cmiqueo, line 1361: This should be `aria-label={title || 'Card Collection'}`

sheridansunier (approved):

cmiqueo (commented):
Shouldn't this PR include the build (dist/*) files?

sanrai:
@cmiqueo thanks for catching two mistakes!

> Shouldn't this PR include the build (dist/*) files?
> This should be aria-label={title || 'Card Collection'}

sheridansunier (approved):

cmiqueo (approved):

PR #188, opened 2024-10-09
feat(mwpw-157946): UI-Tweaks for Newsroom

sheridansunier:
Resolves: [MWPW-157946](https://jira.corp.adobe.com/browse/MWPW-157946)
* adjust height of top filter text to match 'Filter:'


sanrai:
@cmiqueo can you review this?

cmiqueo (approved):

less/components/consonant/filters/top/item.less
```diff
@@ -55,8 +55,7 @@
         &-name {
             display: flex;
             align-items: center;
-
-            .no-margin;
+            margin: 2 0;
```
cmiqueo, line 58: It should be: `margin: 2px 0;`

cmiqueo (changes requested):
Please add a link to the jira ticket for easy cross-referencing :)

less/components/consonant/filters/top/item.less
cmiqueo: You should also add a _Resolves: <Link to jira ticket>_ section for easy cross reference :)

sheridansunier: Both things done @cmiqueo :)

cmiqueo (approved):

PR #211, opened 2024-12-10
fix(MWPW-158561): adds aria-label to search icon

cmiqueo:
Descriiption:
As requested by accessibility team this PR adds the aria-label attribute to search icon

Resolves: 
https://jira.corp.adobe.com/browse/MWPW-158561

Before:
`<button data-testid="search-icon" type="button" class="consonant-SearchIco" tabindex="0" daa-ll="Search-24--Enter a key value pa"><span></span></button>`

After:
`<button data-testid="search-icon" aria-label="Search" type="button" class="consonant-SearchIco" tabindex="0" daa-ll="Search-24--Enter a key value pa"><span></span></button>`


sanrai (approved):

react/src/js/components/Consonant/Search/SearchIcon.jsx
```diff
@@ -25,6 +25,7 @@ const SearchIcon = (props) => {
     return (
         <button
             data-testid="search-icon"
+            aria-label="Search"
```
sanrai, line 28: Need to be localiized.

sanrai (changes requested):

react/src/js/components/Consonant/Search/SearchIcon.jsx
cmiqueo: We are creating a separate ticket to localize all aria-labels

cmiqueo: done!

PR #217, opened 2025-01-13
fix(mwpw-164262): makes headings and text more accessible

cmiqueo:
Description:
As explained in the [WCAG 2.1 website](https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html) the content should be legible with the following text settings:

- Line height (line spacing) to at least 1.5 times the font size;
- Spacing following paragraphs to at least 2 times the font size;
- Letter spacing (tracking) to at least 0.12 times the font size;
- Word spacing to at least 0.16 times the font size.

This PR ensures that if the user locally applies these settings, the card's titles and copy content will still be fully displayed.
The title and content section will expand in height accordingly to accommodate the new text style. 

Resolves: 
https://jira.corp.adobe.com/browse/MWPW-164262


sheridansunier (approved):

.github/workflows/pull-request.yaml
```diff
@@ -98,7 +98,7 @@ jobs:
       - name: Setup Pages
         uses: actions/configure-pages@v2
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v1
+        uses: actions/upload-pages-artifact@v4
```
sanrai, line 101: Should be removed. This may be breaking

sanrai (changes requested):

sanrai:
Will these changes apply and work in Firefox or is that out of scope for this request?

PR #222, opened 2025-01-17
feat(MWPW-164117): adds card title to CTAs aria-label

cmiqueo:
Description: 
Adds card title to CTAs aria-label so that they reader will see something like:
_Read now How to use Photoshop_

Resolves: MWPW-164117


react/src/js/components/Consonant/Infobit/Type/Button.jsx
```diff
@@ -110,6 +113,7 @@ const Button = ({
     const target = getLinkTarget(href, ctaAction);
     const addParams = new URLSearchParams(additionalParams);
     const buttonLink = (additionalParams && addParams.keys().next().value) ? `${href}?${addParams.toString()}` : href;
+    const ariaLabel = `${text} ${title}`;
```
sanrai, line 116: If `title` is empty (which defaults to ``), you end up with a trailing space in the aria-label. 

To fix:
```js
const ariaLabel = title ? `${text} ${title}` : text`
```

react/src/js/components/Consonant/Infobit/Type/Link.jsx
```diff
@@ -31,12 +33,14 @@ const Link = ({
...
     linkHint,
     text,
+    title,
 }) => {
     /**
      **** Authored Configs ****
      */
     const getConfig = useConfig();
     const ctaAction = getConfig('collection', 'ctaAction');
+    const ariaLabel = `${text} ${title}`;
```
sanrai, line 43: If `title` is empty (which defaults to ``), you end up with a trailing space in the aria-label. 

Recommend fixing to:
```js
const ariaLabel = title ? `${text} ${title}` : text`
```

react/src/js/components/Consonant/Infobit/Type/Link.jsx
```diff
@@ -31,12 +33,14 @@ const Link = ({
     href,
     linkHint,
     text,
+    title,
```
sanrai, line 36: Recommend to rename to `linkTitle` as `title` is used below as an html attribute.

sanrai:
## Tests
Verified the following behavior works with secondary ctas

image

sanrai:
###Optional 

Add unit test so we have future regression protection for this feature
```
test('Link includes aria-label combining text and linkTitle', () => {
  render(<Link text="Button text" linkTitle="Some Title" />);
  const link = screen.getByRole('link', { name: /button text some title/i });
  expect(link).toBeInTheDocument();
});
```

sanrai (approved):

react/src/js/components/Consonant/Infobit/Type/Button.jsx
cmiqueo: They do not want to have just the text, it will create duplication for the reader.

PR #226, opened 2025-01-22
fix(mwpw-162902): fix the title sort so that it works with eventFilter

sheridansunier:
Resolves: [MWPW-162902](https://jira.corp.adobe.com/browse/MWPW-162902)

Detailed Description: On the Events Hub page in Milo we have a card collection to show T3 events. Authoring manually hid events that happened in the past. When you sort the cards by A-Z or Z-A, the hidden on-demand cards appear.

................................
Steps to Reproduce:
1. Go to the Events Hub Milo page
2. Scroll down to the card collection at the bottom of the page
3. Sort by A-Z or Z-A, or both

Expected Results: Upcoming event cards are sorted by A-Z or Z-A.
................................
Actual Results: Cards are sorted and On-Demand cards display.


sanrai (approved):

sanrai:
After the PR is merged, we should manually test the following scenarios (and have some test pages set up for this)

### Test Scenarios

#### A. No eventFilter + Title Sort

- Remove any eventFilter.
- Set sort to “titleAsc” (or “titleDesc,” whichever you want to test).
- Verify that the cards appear in the correct title order (A → Z for Asc, Z → A for Desc).
- This ensures the normal title sort is still working as expected in the absence of an event filter.

#### B. With eventFilter + Title Sort

- Apply an eventFilter that is expected to narrow down your cards to a subset.
- Set sort to “titleAsc” (or “titleDesc”).
- Verify that only the cards matching the eventFilter are shown.
- Verify that those filtered cards are ordered correctly by title.
    - For “Asc,” the titles should be in ascending alphabetical order.
    - For “Desc,” the titles should be in descending alphabetical order.

#### C. With eventFilter + Other Sorts (Optional but Recommended)

- Apply the eventFilter as in the previous step.
- Set the sort to something else, like “dateAsc” or “dateDesc.”
- Verify that only the filtered cards appear and they are sorted by date ascending/descending correctly.

#### D. Edge Cases

- Empty eventFilter: Confirm that if the filter is empty, no special filtering is applied, and sort still works normally.
- No Cards Matching: If the filter is set so that no cards match, make sure the UI handles that gracefully (e.g., zero results).
- Featured Cards / Hidden CTA:For featured cards or for hiding CTAs, verify that these features still work the same way after filter and sort changes.

PR #227, opened 2025-01-23
fix(mwpw-166139): issue with footers with 0.25.0

sheridansunier:
When working on the newsroom card style story - I noticed that for certain card styles, the new changes were breaking card collections.  This fixes that bug.

**BEFORE:**
Screen Shot 2025-01-23 at 2 22 19 PM

**AFTER:**
Screen Shot 2025-01-23 at 2 26 19 PM


react/src/js/components/Consonant/Helpers/Helpers.js
```diff
@@ -413,8 +413,13 @@ export const getDateDescSort = cards => getDateAscSort(cards).reverse();
 export const getEventSort = (cards = [], eventFilter) => {
     const transformedCards = cards.map(card => ({
         id: card.id,
-        startDate: card.contentArea.dateDetailText.startTime || card.footer[0].left[1].startTime || '',
-        endDate: card.contentArea.dateDetailText.endTime || card.footer[0].left[1].endTime || '',
+        endDate: card.contentArea.dateDetailText.endTime || '',
+        ...(card.footer[0].left[1] ?
```
sanrai, line 417: If `card.footer` is not defined or an empty array, then `card.footer[0].left[1]` will cause a runtime error.

react/src/js/components/Consonant/Helpers/Helpers.js
```diff
@@ -413,8 +413,13 @@ export const getDateDescSort = cards => getDateAscSort(cards).reverse();
 export const getEventSort = (cards = [], eventFilter) => {
     const transformedCards = cards.map(card => ({
         id: card.id,
-        startDate: card.contentArea.dateDetailText.startTime || card.footer[0].left[1].startTime || '',
-        endDate: card.contentArea.dateDetailText.endTime || card.footer[0].left[1].endTime || '',
+        endDate: card.contentArea.dateDetailText.endTime || '',
```
sanrai, line 416: This code is double assigning`endDate`. 

Imagine if the card is defined as this:

```
const card = {
  contentArea: {
    dateDetailText: {
      startTime: '2023-10-01',
      endTime:   '2023-10-02'
    }
  },
  footer: [{
    left: [
      // 0: could be something else
      {
        startTime: '2024-01-01',
        endTime:   '2024-01-02'
      }
    ]
  }]
};
```

The code is essentially doing this:
```
endDate: card.contentArea.dateDetailText.endTime || '',

...(
  card.footer[0].left[1]
    ? { endDate: card.footer[0].left[1].endTime }
    : { endDate: card.contentArea.dateDetailText.endTime }
),
```

**Step by step:**
1. First assignment
```
endDate = '2023-10-02' // taken from card.contentArea.dateDetailText.endTime
```

2. Second assignment

Because `card.footer[0].left[1]` exists, the ternary picks

```
endDate = '2024-01-02'
```

from the footer. 

This overwrites `2023-10-02`.

Effectively, the second assignment always wins if `footer[0].left[1]` is present. The first line of code runs, but then immediately gets replaced. 

Is this expected?

react/src/js/components/Consonant/Helpers/__tests__/CardFilterer.spec.js
```diff
@@ -79,12 +79,12 @@ describe('utils/CardFilterer', () => {
 
         test('Event Sort', () => {
             const cards = [
-                { id: 1, contentArea: { dateDetailText: { startTime: '2021-01-01', endTime: '2021-01-02' } }, tags: ['event1'] },
-                { id: 2, contentArea: { dateDetailText: { startTime: '2021-01-03', endTime: '2029-01-04' } }, tags: ['event2'] },
+                { id: 1, contentArea: { dateDetailText: { startTime: '2021-01-01', endTime: '2021-01-02' } }, tags: ['event1'], footer: [{ left: []}] },
```
sanrai, line 82: We should add test coverage for our fallback logic for `startDate/endDate` when `card.footer[0].left[1]` may or may not exist, so we don’t throw errors or pick the wrong date. 

The key scenarios:

1. No `footer` at all – Confirm we gracefully use `contentArea.dateDetailText`.
2. `footer` exists but only [0] in left – Ensure no error and fallback works.
3. `footer[0].left[1]` fully present – Confirm it overrides contentArea.
4. Partial `footer[0].left[1]` – If one date is missing, ensure fallback grabs it from contentArea.

sanrai (changes requested):

react/src/js/components/Consonant/Helpers/Helpers.js
sheridansunier: so my investigation into our data/response from chimera-io... card.footer[0].left never doesn't come back - the reason I made this change was to compensate for the fact that card.footer[0].left[1] sometimes does not exist (which was causing the runtime error you are anticipating) however I totally am writing endDate twice - that was a mistake I left in lol

react/src/js/components/Consonant/Helpers/Helpers.js
sheridansunier: not going to leave enddate twice - that was not done on purpose/originally existed so I could test stuff

react/src/js/components/Consonant/Helpers/Helpers.js
sanrai: ​​Even if your data typically includes `footer[0].left`, you still risk a runtime error if footer is empty or left doesn’t have index.

sanrai (changes requested):

sanrai:
Right now, we're doing:

```
...((card.footer && card.footer[0].left.length > 1) ?
    { startDate: card.footer[0].left[1].startTime } :
    { startDate: card.contentArea.dateDetailText.startTime }),
```

If `card.footer` is an empty array, then `card.footer[0]` is `undefined`, causing a runtime error when we do `card.footer[0].left`.

If your data always includes `contentArea.dateDetailText`, you're safe. 

Otherwise, something like `card.contentArea.dateDetailText.startTime` might also throw an error if `contentArea` or `dateDetailText` is missing.

We can avoid these edge-case errors by adding defensive checks (e.g., multiple if statements) or by using a helper function like this:

```
/**
 * Convert a path string like 'footer[0].left[1].startTime'
 * into an array of keys: ['footer','0','left','1','startTime'].
 */
function parsePathString(pathString) {
  return pathString
    // Convert bracket notation to dot notation
    // e.g. 'footer[0].left[1]' → 'footer.0.left.1'
    .replace(/\[(\d+)\]/g, '.$1')
    // Now split on '.'
    .split('.');
}

/**
 * Safely get a nested property from an object 
 * using a path string with dot/bracket notation.
 * e.g. safeGet(card, 'footer[0].left[1].startTime', '')
 */
function safeGet(obj, pathString, defaultVal) {
  const parts = parsePathString(pathString);
  let current = obj;
  
  for (let i = 0; i < parts.length; i++) {
    if (current == null || typeof current !== 'object') {
      return defaultVal;
    }
    const key = parts[i];
    if (!(key in current)) {
      return defaultVal;
    }
    current = current[key];
  }

  return current == null ? defaultVal : current;
}

// ----------------------------------------
// Usage example:
// ----------------------------------------

const card = {
  footer: [
    {
      left: [
        { type: 'bookmark' },
        { type: 'date-interval', startTime: '2023-01-01T10:00:00Z' }
      ]
    }
  ]
};

// This will return '2023-01-01T10:00:00Z'
const dateValue = safeGet(card, 'footer[0].left[1].startTime', '');
console.log(dateValue); // '2023-01-01T10:00:00Z'

// If 'footer[0].left[2].startTime' doesn’t exist,
// we get the default value '' instead of throwing an error.
const nonExistentValue = safeGet(card, 'footer[0].left[2].startTime', '');
console.log(nonExistentValue); // ''
```

Using `safeGet` (or similar guard logic) helps prevent runtime errors if parts of the object path are missing.

sanrai (changes requested):
Right now, we're doing:

```
...((card.footer && card.footer[0].left.length > 1) ?
    { startDate: card.footer[0].left[1].startTime } :
    { startDate: card.contentArea.dateDetailText.startTime }),
```

If `card.footer` is an empty array, then `card.footer[0]` is `undefined`, causing a runtime error when we do `card.footer[0].left`.

If your data always includes `contentArea.dateDetailText`, you're safe. 

Otherwise, something like `card.contentArea.dateDetailText.startTime` might also throw an error if `contentArea` or `dateDetailText` is missing.

We can avoid these edge-case errors by using a helper function like this:

```
/**
 * Convert a path string like 'footer[0].left[1].startTime'
 * into an array of keys: ['footer','0','left','1','startTime'].
 */
function parsePathString(pathString) {
  return pathString
    // Convert bracket notation to dot notation
    // e.g. 'footer[0].left[1]' → 'footer.0.left.1'
    .replace(/\[(\d+)\]/g, '.$1')
    // Now split on '.'
    .split('.');
}

/**
 * Safely get a nested property from an object 
 * using a path string with dot/bracket notation.
 * e.g. safeGet(card, 'footer[0].left[1].startTime', '')
 */
function safeGet(obj, pathString, defaultVal) {
  const parts = parsePathString(pathString);
  let current = obj;
  
  for (let i = 0; i < parts.length; i++) {
    if (current == null || typeof current !== 'object') {
      return defaultVal;
    }
    const key = parts[i];
    if (!(key in current)) {
      return defaultVal;
    }
    current = current[key];
  }

  return current == null ? defaultVal : current;
}

// ----------------------------------------
// Usage example:
// ----------------------------------------

const card = {
  footer: [
    {
      left: [
        { type: 'bookmark' },
        { type: 'date-interval', startTime: '2023-01-01T10:00:00Z' }
      ]
    }
  ]
};

// This will return '2023-01-01T10:00:00Z'
const dateValue = safeGet(card, 'footer[0].left[1].startTime', '');
console.log(dateValue); // '2023-01-01T10:00:00Z'

// If 'footer[0].left[2].startTime' doesn’t exist,
// we get the default value '' instead of throwing an error.
const nonExistentValue = safeGet(card, 'footer[0].left[2].startTime', '');
console.log(nonExistentValue); // ''
```

Using `safeGet` (or similar guard logic) helps prevent runtime errors if parts of the object path are missing.

react/src/js/components/Consonant/Helpers/Helpers.js
```diff
@@ -413,8 +447,10 @@ export const getDateDescSort = cards => getDateAscSort(cards).reverse();
 export const getEventSort = (cards = [], eventFilter) => {
     const transformedCards = cards.map(card => ({
         id: card.id,
-        startDate: card.contentArea.dateDetailText.startTime || card.footer[0].left[1].startTime || '',
-        endDate: card.contentArea.dateDetailText.endTime || card.footer[0].left[1].endTime || '',
+        startDate: safeGet(card, 'contentArea.dateDetailText.startTime', safeGet(card, 'footer[0].left[1].startTime', '')),
+        ...((card.footer && card.footer[0].left.length > 1) ?
```
sanrai, line 451: Will throw an exception if `card.footer` is `[]`. It's not catching/checking for this. 

Recommend swapping to `safeGet` to never worry about this.

react/src/js/components/Consonant/Helpers/Helpers.js
```diff
@@ -413,8 +447,10 @@ export const getDateDescSort = cards => getDateAscSort(cards).reverse();
 export const getEventSort = (cards = [], eventFilter) => {
     const transformedCards = cards.map(card => ({
         id: card.id,
-        startDate: card.contentArea.dateDetailText.startTime || card.footer[0].left[1].startTime || '',
-        endDate: card.contentArea.dateDetailText.endTime || card.footer[0].left[1].endTime || '',
+        startDate: safeGet(card, 'contentArea.dateDetailText.startTime', safeGet(card, 'footer[0].left[1].startTime', '')),
```
sanrai, line 450: Before you had, “use footer if it exists, otherwise fallback to contentArea.” The latest update has this swapped.

react/src/js/components/Consonant/Helpers/Helpers.js
```diff
@@ -413,8 +447,10 @@ export const getDateDescSort = cards => getDateAscSort(cards).reverse();
...
+        ...((card.footer && card.footer[0].left.length > 1) ?
+            { endDate: card.footer[0].left[1].endTime } :
+            { endDate: card.contentArea.dateDetailText.endTime }),
```
sanrai, line 453: Would recommend updating `endDate` to use `safeGet` to avoid bug/issue above with ternary/spread (with `[]` case)

sanrai (changes requested):
Recommended fix/snippet:

```
startDate: safeGet(
  card, 
  'footer[0].left[1].startTime', 
  safeGet(card, 'contentArea.dateDetailText.startTime', '')
),
endDate: safeGet(
  card, 
  'footer[0].left[1].endTime', 
  safeGet(card, 'contentArea.dateDetailText.endTime', '')
),
```

react/src/js/components/Consonant/Helpers/Helpers.js
sheridansunier: I think I accidentally left the endDate there lol - I'll swap it to match startDate

react/src/js/components/Consonant/Helpers/Helpers.js
sheridansunier: I think I would rather have it go contentArea then footer -  first was a first attempt at avoiding the runtime error - almost all of our datastream have contentArea and only the events stream uses footer

sanrai: Yeah that's fine :) 

I just wanted to note it in case it was unintended

sanrai (approved):

PR #232, opened 2025-01-29
fix(mwpw-166139): fixing event sort

sheridansunier:
**Before:**
Screen Shot 2025-01-29 at 2 08 29 PM

**After:**
Screen Shot 2025-01-29 at 2 08 55 PM


sanrai (approved):
:)

PR #258, opened 2025-03-11
fix(mwpw-169194): sanitizes createdDate format used by the detailsText

cmiqueo:
Description:
Sanitizes createdDate format when the timestamp is set to 00:00:00.0000Z 

Simple and effective implementation that when the timestamp is set to  **00:00:00.0000Z** removes the '**Z**' part of the string before creating the JavaScript object and using the local timezone instead of 'GMT'

Test page:
[Content as a Service v3 - Tuesday, March 11, 2025 at 10:15](https://milo.stage.adobe.com/tools/caas#~~H4sIAAAAAAAAExXMMQoCMRAF0Lv8Oo1aJa32Lmw6sRiSQRfCTphMWIN4d0n7ivdFIs2rjcIIOIUzHAoN6RZHnXTpFQ7KrRdrC+tCr8keDk26JkZ4YOej4emQ2WgrLfLH7tU22RGQlMk438gYDiZG5UqaW5T1LcesvMfvD+c+F+iIAAAA)


sheridansunier (approved):

sanrai (approved):

react/src/js/components/Consonant/Cards/Card.jsx
```diff
@@ -210,6 +210,9 @@ const Card = (props) => {
     } else if (detailsTextOption === 'createdDate' && cardDate) {
         const localCreatedDate = new Date(cardDate);
         detailText = localCreatedDate.toLocaleDateString();
+    } else if (detailsTextOption === 'staticDate' && cardDate) {
+        const staticDate = new Date(cardDate.replace(/Z$/, ''));
```
sanrai, line 214: This is a very clean solution!

Much better than just splitting by `T` and taking the left hand (as it'll also localize the dates in the right format)

sanrai (approved):

PR #267, opened 2025-04-15
fix(mwpw-169861): properly chaining promises before calling alloy 

shkhan91:
#### Functional Updates

- Updated `Container.jsx` to:
  - Use `window.alloy_getIdentity` as the **preferred identity provider**.
  - Fallback to `window.alloy('getIdentity')` **after** `window._satellite.alloyConfigurePromise` resolves.
  - If both Alloy APIs fail or are unavailable, gracefully fallback to the Visitor API.
  - If Visitor API also fails, fall back to calling the base `collectionEndpoint` without any targeting data.
- Ensured `currentEntityId` is appended to requests if `useLastViewedSession` is enabled and a value is stored in `localStorage`.

#### Testing Enhancements

- Added new test cases in `Alloy.spec.js`:
  - Confirm `window.alloy_getIdentity` is used when available.
  - Verify fallback to `window.alloy('getIdentity')` if `alloy_getIdentity` is unavailable or fails.
  - Verify fallback to Visitor API when Alloy fails or returns incomplete identity info.
  - Validate default endpoint call when all identity strategies fail.
  - Validate `currentEntityId` injection from `localStorage` when enabled.
- Covered previously untested fallback branches to increase overall function coverage.

---

### Jira References

- `MWPW-169861` – Resolve timing issues with satellite configuration
- `ENB-7928` – Improve resiliency of identity resolution and Target fallback handling

---

### Outcome

This PR ensures identity resolution is:
- Safe and layered (tries `alloy_getIdentity`, then `alloy('getIdentity')`, then Visitor API)
- Resilient to missing configs, malformed identities, and promise timing issues


react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -941,42 +943,77 @@ const Container = (props) => {
...
-            });
+        function alloyApiFallback(visitor, url) {
+            const satellite = window._satellite; // eslint-disable-line no-underscore-dangle
+            if (window.alloy && window.edgeConfigId && satellite
+                && satellite.alloyConfigurePromise) {
+                satellite.alloyConfigurePromise
+                    .then(() => window.alloy('getIdentity'))
+                    .then((res) => {
+                        if (
+                            res && res.identity && res.identity.ECID &&
```
sanrai, line 968: Instead of having all this nesting, you can just use `safeGet` 

For example: `safeGet(res, 'identidy.ECID', false)`. 

```
import {
    safeGet // Add this import
} from '../Helpers/Helpers';
```

sanrai (changes requested):

sanrai (approved):
Instead of having all this nesting, you can just use `safeGet` 

For example: `safeGet(res, 'identidy.ECID', false)`. 

```
import {
    safeGet // Add this import
} from '../Helpers/Helpers';
```

sheridansunier (approved):

sanrai:
So right now it looks like the code does this:

```
Visitor Promise fails? → getCards(collectionEndpoint) → on error, fallbackEndpoint
Visitor Promise → Alloy/Visitor succeeds? → getCards(urlWithParams) → on error, no static fallback
Visitor Promise → Alloy/Visitor fails late? → visitorApiFallback → getCards(urlWithParams) → on error, no static fallback
```

What we probably intend is:

    A) First try the Adobe/Visitor‑augmented URL
    B) If that fetch fails, drop back to the plain collectionEndpoint
    C) If that fetch fails, drop back to fallbackEndpoint

Please double check this:

1. `getCards` built‑in fallback only fires when we fetch the un‑parameterized collectionEndpoint.   In our `getCards(endPoint)` we have:
```js
               .catch(() => {
                 if (endPoint === collectionEndpoint && fallbackEndpoint) {
                   // only when they’re exactly the same string
                   getCards(fallbackEndpoint);
                   return;
                 }
                 // log error + setApiFailure
               });
```
That means if we called `getCards(collectionEndpoint)` (or simply `getCards()`), and that fetch bombs, we'll then try our Akamai‑hosted JSON.
           
2. Identity‐enhanced calls bypass that fallback completely. In our new Adobe/Visitor logic we end up doing things like
```js
getCards(urlWithMcgvidAndOtherParams);
```
 If *that* fetch fails, our `.catch()` sees `endPoint !== collectionEndpoint` and so it just logs the error & sets `apiFailure`—it never tries the`fallbackEndpoint`.

sanrai:
The switch from using `window.alloy('getIdentity')` to `window.allow_getIdentity` doesn't look like it resolves correctly on prod (`regionId` isn't captured).

image

sanrai:
Overall your new tests do a great job of covering the various Alloy vs. Visitor‐API code paths and the “reveal ID” logic!

Here are a couple of ideas of higher‑level gaps and opportunities to tighten up:

1. Missing coverage of the static‑JSON (fallbackEndpoint) chain
• There are no tests that simulate a fetch rejection on the identity URL, then a rejection on the plain collection URL, followed by a successful fetch of your Akamai JSON.

2. Explicitly testing the plain→static fallback when the Visitor API is disabled or times out
• If `target.enabled: false` (or no `__satelliteLoadedPromise`) we should still fall back to static JSON on fetch failure.
• Add a test where we clear out `window.__satelliteLoadedPromise` and simulate a fetch rejection, then verify a static‐file call.

3. Empty‐payload (no cards) branch
 • In `getCards` we have a check for “no valid response data” that rejects.
 • A test that returns `{}` or `{ cards: [] }` and verifies we log an error and don’t crash would be good.

shkhan91:
> The switch from using `window.alloy('getIdentity')` to `window.allow_getIdentity` doesn't look like it resolves correctly on prod (`regionId` isn't captured).

Yes, I spoke with the team that implemented the Alloy API. They want us to use the window.allow_getIdentity, which is faster and loads quicker. I told them to add the regionID to that API, so it's usable for us. For now we will be falling back until they make changes on their end.

cmiqueo (approved):

PR #268, opened 2025-04-16
feat(mwpw-164509): fixes carousel navigation issues

cmiqueo:
Description: 
Refactors carousel for improved accessibility

Resolves:
https://jira.corp.adobe.com/browse/MWPW-164509

Test page:
https://milo.adobe.com/tools/caas?caasver=alpha


react/src/js/components/Consonant/Filters/Top/__tests__/Group.spec.js
```diff
@@ -0,0 +1,52 @@
...
+        
+        // First click - should change to "Close"
+        fireEvent.click(filterButton);
+        expect(filterNameElement).toHaveAttribute('daa-ll', `${name} Close`);
+        
+        // Second click - should change back to "Open"
+        fireEvent.click(filterButton);
+        expect(filterNameElement).toHaveAttribute('daa-ll', `${name} Open`);
+    });
+});
```
shkhan91, line 52: Nit pick please add EOL

shkhan91 (approved):

react/src/js/components/Consonant/Cards/Card.jsx
```diff
@@ -552,23 +555,26 @@ const Card = (props) => {
...
+                        renderOverlay={renderOverlay} />
                 ))}
                 {(isThreeFourths || isDoubleWide || isFull)
                     && !renderOverlay
                     && <LinkBlocker
                         target={linkBlockerTarget}
                         link={overlay}
                         title={title}
-                        getsFocus={getsFocus}
+                        getsFocus={getsFocus || true}
```
sanrai, line 568: If this is always true, doesn't that completely defeat the purpose of “only focus certain cards”?

react/src/js/components/Consonant/Infobit/Type/Button.jsx
```diff
@@ -22,6 +22,8 @@ const buttonType = {
     isCta: bool,
     onFocus: func,
     title: string,
+    tabIndex: string,
```
sanrai, line 25: PropTypes declares `tabIndex: string` with a default of `''`, but DOM `tabIndex` is almost always numeric (0 or –1). 

Not sure we should have an empty-string fallback for this (since I think this will create invalid HTML)

Ideally PropType would be `number` and default to `0`.

react/src/js/components/Consonant/Cards/Card.jsx
```diff
@@ -399,7 +402,7 @@ const Card = (props) => {
     const altCtaLink = getAltCtaLink(footer);
     const ctaText = (altCtaUsed && isUpcoming && altCtaLink !== '') ? getCtaText(footer, 'alt') : getCtaText(footer, 'right');
     const overlay = (altCtaUsed && isLive && altCtaLink !== '') ? altCtaLink : overlayParams;
-    const getsFocus = (isHalfHeight && !videoURLToUse)
+    const getsFocus = isHalfHeight
```
sanrai, line 405: Since we removed `&& !videoURLToUse` just double checking that cards with videos shouldn’t be focusable?

react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
```diff
@@ -115,63 +176,80 @@ function CardsCarousel({
...
-    function scrollHandler() {
-        interactedWith = true;
-        responsiveLogic();
+            if (index + 1 >= firstVisibleCard && index + 1 <= lastVisibleCard) {
+                cardLink.removeAttribute('aria-hidden');
+                cardLink.removeAttribute('inert');
+                cardLink.setAttribute('tabindex', '0');
+            } else {
+                cardLink.setAttribute('aria-hidden', 'true');
+                cardLink.setAttribute('inert', '');
```
sanrai, line 210: HTML `inert` attribute is not universally supported (e.g. Safari). Do we need to ship a polyfill as well?

react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
```diff
@@ -46,36 +98,35 @@ function CardsCarousel({
     const prev = useRef(null);
     const next = useRef(null);
 
-    let isDown = null;
-    let startX = null;
-    /* eslint-disable-next-line no-unused-vars */
-    let isMouseMove = false;
-    let interactedWith = false;
+    let firstVisibleCard = 1;
```
sanrai, line 101: `firstVisibleCard`, `lastVisibleCard` should be hoisted into refs, 

```
const firstVisibleCard = useRef(1);
```

Otherwise these locals get re‑initialized on every render.

react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
```diff
@@ -115,63 +176,80 @@ function CardsCarousel({
...
 
     function nextButtonClick() {
-        if (isResponsive()) {
+        if (isMobile()) {
             centerClick();
         } else {
             const carousel = carouselRef.current;
             carousel.scrollLeft += ((cardWidth + gridGap) * cardsShiftedPerClick);
+            setVisibleCards('next');
+            setAriaAttributes(carousel);
```
sanrai, line 239: The `setAriaAttributes(...)` only get called in the next/prev handlers—so on initial render none of the off‑screen cards get `aria-hidden` or `tabindex=-1`

You may want to move an initial call to `setAriaAttributes(carouselRef.current)` into your `useEffect`.

react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
```diff
@@ -189,7 +267,27 @@ function CardsCarousel({
     const totalResultsHtml = RenderTotalResults(showTotalResultsText, resQty);
 
     useEffect(() => {
-        responsiveLogic();
+        mobileLogic();
+
+        const carousels = document.querySelectorAll('.consonant-Container--carousel');
```
sanrai, line 272: Instead of doing a query, you may want to try using a ref or at least memoize the card list so you’re not searching the entire DOM on every time.

react/src/js/components/Consonant/Grid/Grid.jsx
```diff
@@ -239,8 +257,10 @@ const Grid = (props) => {
                                 locale={locale}
                                 renderBorder={renderCardsBorders}
                                 renderDivider={renderFooterDivider}
-                                renderOverlay={renderCardsOverlay}
+                                renderOverlay={renderOverlay}
                                 hideCTA={hideCTA}
+                                ariaHidden={ariaHidden}
+                                tabIndex={ariaHidden ? '-1' : ''}
```
sanrai, line 263: Returning `` should create invalid HTML.

react/src/js/components/Consonant/Testing/Utils/MockCards.js
```diff
@@ -0,0 +1,25 @@
+import { DEFAULT_PROPS_ONE_HALF } from '../Constants/Card';
```
sanrai, line 1: Was this file intended to be committed? Its not being referenced/used anywhere.

sanrai (approved):

PR #288, opened 2025-06-04
test(mwpw-1234): saving wip

raissanjay:


pa11y.json
```diff
@@ -1,10 +1,14 @@
-{ "chromeLaunchConfig":
-  {
+{
+  "delay": 20000,
```
raissanjay, line 2: To ensure caas fully loads since it connects to chimera and does full e2e.

PR #297, opened 2025-06-25
fix(MWPW-175459): reads next and prev buttons from i18n

cmiqueo:
Description:
Reads "Next Cards" and "Previous Cards" aria-label sctring from **collection.i18n** 

Resolves:
https://jira.corp.adobe.com/browse/MWPW-175459


sanrai:
Just a note. This PR:
https://github.com/adobecom/milo/pull

Needs to be merged before this one in CaaS can be.

PR #315, opened 2025-08-21
fix(mwpw-1234): upgrade

sanrai:
placeholder text


sanrai:
Those comments come from linting rules as warnings but we use these APIs such as `dangerouslySetInnerHTML` to handle various business requirements (since there is no other way).

PR #365, opened 2025-12-10
feat(mwpw-183491): implements modern carousel style

cmiqueo:
**Description**
This PR adds a new feature to the CaaS Carousel layout.

It allows the user to choose between two different navigation control styles:
- Modern design
- Classic design

The modern design is a new design the latest version of the CaaS Carousel navigation controls.
The classic design remains unchanged from the original carousel navigation controls design.

THe user will need to select the new Modern design in the CaaS Configuration Tool.
When Carousel layout option is selected, the user will see a drop down with the following options:
- Modern (Paged)
- Modern (Incremental)
- Classic (Paged)
- Classic (Incremental)

The Paged option advances the carousel by one page at a time.
The Incremental option advances the carousel by one card at a time.

There is also a checkbox option to use a light background for the navigation controls. 
 [x] Light background

This will affect only the Modern design as specified in the Figma files

**Resolves** 
https://jira.corp.adobe.com/browse/MWPW-183491

### DIFF
--- a/react/src/js/components/Consonant/Container/Container.jsx
+++ b/react/src/js/components/Consonant/Container/Container.jsx
@@ -7,7 +7,7 @@ import React, {
 } from 'react';
 import classNames from 'classnames';
 import { shape } from 'prop-types';
-import 'whatwg-fetch';
+// import 'whatwg-fetch'; // Removed: fetch is native in modern browsers
 import { logLana } from '../Helpers/lana';
 import Popup from '../Sort/Popup';
 import Search from '../Search/Search';
@@ -20,6 +20,7 @@ import {
     sanitizeEventFilter,
     getTransitions,
     removeMarkDown,
+    preloadFirstCardImage,
 } from '../Helpers/general';
 import { configType } from '../types/config';
 import CardsCarousel from '../CardsCarousel/CardsCarousel';
@@ -66,6 +67,8 @@ import {
     getActiveFilterIds,
     getActivePanels,
     getUpdatedCardBookmarkData,
+    transformFiltersWithCategories,
+    expandGroupFiltersToChildren,
 } from '../Helpers/Helpers';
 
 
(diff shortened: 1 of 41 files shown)


react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
```diff
@@ -113,33 +114,49 @@ function CardsCarousel({
 
     function hideNextButton() {
         const nextBtn = next.current;
-        if (nextBtn) {
+        if (nextBtn && carouselType === 'default') {
             nextBtn.classList.add('hide');
             nextBtn.setAttribute('aria-hidden', 'true');
+        } else {
```
raissanjay: What if `nextBtn` is null? The else block will throw an error.  Need `else if` check.

Same issue in `hidePrevButton()`, `showNextButton()`, and `showPrevButton()`.

react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -163,6 +160,11 @@ const Container = (props) => {
     const partialLoadWithBackgroundFetch = getConfig('collection', 'partialLoadWithBackgroundFetch.enabled');
     const partialLoadCount = getConfig('collection', 'partialLoadWithBackgroundFetch.partialLoadCount');
     const renderOverlay = getConfig('collection', 'useOverlayLinks');
+    // const isModernCarousel = getConfig('pagination', 'animationStyle')?.toLowerCase().includes('modern');
```
raissanjay: Should this comment be removed?

raissanjay (approved):

sanrai (approved):

sanrai (changes requested):

react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
cmiqueo: @sanrai, hey! good catch!!!

sanrai (approved):

PR #451, opened 2026-04-30
fix(mwpw-177207): adds visible Sort by: label to sort dropdown

cmiqueo:
**Description**:
Adds a visible text label before the sort dropdown button in the Consonant top-filter panel, replacing the implicit-only aria-label with an explicit paired <label> element.

**L10n:**
It can be localized by using the following new key in the mappings file 
`sortBy: "Sort by:"`

**Resolves**:
https://jira.corp.adobe.com/browse/MWPW-177207

**Before**:
Screenshot 2026-04-30 at 8 41 40 AM

**After**:
Screenshot 2026-04-30 at 8 41 12 AM

### DIFF
--- a/less/components/consonant/filters/top/panel.less
+++ b/less/components/consonant/filters/top/panel.less
@@ -13,7 +13,7 @@
         &-inner {
             display: flex;
             flex-wrap: wrap;
-            justify-content: space-between;
+            justify-content: flex-end;
             align-items: flex-start;
             max-width: 100%;
         }
@@ -37,7 +37,7 @@
             justify-content: flex-end;
             align-items: flex-start;
             flex-grow: 1;
-            max-width: calc(~ '100% - ' @consonant-Select-width * 0.85 ~ ' - 8px');
+            max-width: calc(~ '100% - ' @consonant-Select-width ~ ' - 8px');
 
             &:first-child {
                 max-width: 100%;
@@ -157,7 +157,6 @@
         &-selectWrapper {
             display: flex;
             justify-content: flex-end;
-            width: @consonant-Select-width * 0.85;
             margin-right: 8px;
             order: -1;
             overflow: visible;
@@ -273,7 +272,6 @@
             }
 
             &-selectWrapper {
-                max-width: @consonant-Select-width;
                 order: 0;
                 margin-right: 0;
             }
@@ -316,13 +314,12 @@
 
             &-selectWrapper {
                 min-width: auto;
-                max-width: 160px;
                 margin-right: 0;
                 margin-left: 0;
 
                 & > div:first-child > button {
-                    padding-left: 12px;
-                    padding-right: 12px;
+                    // padding-left: 12px;
+                    padding-right: 24px;
                     border: 1px solid @consonantLightGrey200;
                     border-radius: 4px;
                 }
@@ -331,7 +328,7 @@
 
                 & > div:first-child {
                     &:after {
-                        right: 12px;
+                        right: 4px;
                     }
                 }
             }
--- a/less/components/consonant/filters/top/panel.less
+++ b/less/components/consonant/filters/top/panel.less
@@ -318,7 +318,6 @@
                 margin-left: 0;
 
                 & > div:first-child > button {
-                    // padding-left: 12px;
                     padding-right: 24px;
                     border: 1px solid @consonantLightGrey200;
                     border-radius: 4px;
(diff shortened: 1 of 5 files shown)


shkhan91 (approved):

less/components/consonant/filters/top/panel.less
```diff
@@ -316,13 +314,12 @@
...
             &-selectWrapper {
                 min-width: auto;
-                max-width: 160px;
                 margin-right: 0;
                 margin-left: 0;
 
                 & > div:first-child > button {
-                    padding-left: 12px;
-                    padding-right: 12px;
+                    // padding-left: 12px;
```
raissanjay: Worth deleting outright rather than leaving it comented out. Lints will skip it but the next person edting this rule will wonder whether 12px was meant tocome back. If the intent is to preserve the originalvalue for reference, a one-line comment above explaiing why (e.g. "intentionally no left padding, label its to the left") would carry more signal than the lteral old line.

react/src/js/components/Consonant/Sort/__tests__/Popup.spec.js
```diff
@@ -9,6 +9,14 @@ import { testAccessibility } from '../../Testing/Utils/a11yTest';
 const renderSortPopup = setup(Popup, DEFAULT_PROPS);
 
 describe('Consonant/Sort/Popup', () => {
+    test('Should show sort by label', async () => {
+        renderSortPopup();
+        const sortPopLabel = screen.getByTestId('consonant-Select-label');
+
+        expect(sortPopLabel).toHaveTextContent('Sort by:');
+        expect(sortPopLabel).not.toHaveAttribute('aria-label');
```
raissanjay, line 17: The `not.toHaveAttribute('aria-label')` assertion is the right anti-regression check. Without it, a future patch could easily reintroduce an aria-label thinking it adds value, and screen readers would then announce the same control twice.

raissanjay (approved):
**Tested locally**

- Built the PR branch (`NODE_ENV=production npx webpack`) — emitted `dist/main.js` (890 KB) and `dist/app.css` (227 KB), no warnings
- Confirmed bundle contains the new strings: `consonant-Select-label`, `sort-by-Select`, `Sort by:`
- Served the built dist via WebStorm's built-in server (`localhost:63342/caas/`), compared before/after on the same test page
- DOM checks on the rendered after state:
  - `[data-testid="consonant-Select-label"]` element renders
  - Label text content = `"Sort by:"`
  - Label `htmlFor` = `"consonant-Select-btn"`
  - Button `id` = `"consonant-Select-btn"` (matches `htmlFor`)
  - Label has no `aria-label` attribute
  - Wrapper carries both `consonant-Select` and `sort-by-Select` classes
  - Wrapper computed `display: flex`, label computed `font-size: 13px`
  - Gap between label end and button start measured at 0 px (baseline alignment works)
- Dropdown still functional: click opens popup with all 6 sort options (Random, Featured, Title A-Z, Title Z-A, Date Oldest, Date Newest), `aria-expanded` toggles to `"true"`
- New test file syntax-parses without errors (`node --check`)
- **Top-filter panel layout (`filterPanel.type: 'top'`)**: switched test page config, verified the layout reflows correctly. `.consonant-TopFilters` renders, sort widget sits at far right per the new `justify-content: flex-end` (replacing `space-between`), wrapper sized to fit just label + button (137 px width), no orphan whitespace. The CSS changes inside `@media @consonant-desktop-up` apply cleanly.
- **Narrow viewport (375 px)**: constrained body to 375 px via injected style and reloaded — `Sort by:` label and button remain adjacent on a single line with baseline alignment intact, no wrap or overlap. (Note: the `panel.less` width-removal changes are inside `@media @consonant-desktop-up` — the `select.less` `.sort-by-Select` rule applies at all widths.)
- **i18n override path**: injected `i18n: { sortBy: 'Trier par : (custom)' }` into the test page's collection config; label rendered the custom string verbatim. Confirms `getConfig('collection', 'i18n.sortBy')` reads through the override correctly. Authors can swap the label text without touching React code.
- **RTL layout** (`<html dir="rtl" lang="ar">`): the whole page mirrors as expected — filter panel moves to the right, cards reorder, and the sort widget moves to the visual top-left (mirror of LTR top-right). Inside the widget, `display: flex` flips the label/button visual order so `Sort by:` appears on the visual right and the dropdown on the visual left, reading naturally in RTL order. Colon position is correctly handled by the Unicode bidi algorithm for English text in an RTL context (would render properly in actual Arabic).

Approving. Verified items look clean across desktop, top-filter layout, narrow viewport, i18n override, and RTL. The only thing I'd nudge before merge is the dead commented-out line in panel.less (inline comment above).

shkhan91 (approved):

react/src/js/components/Consonant/Sort/__tests__/Popup.spec.js
cmiqueo: done

PR #453, opened 2026-05-07
feat(mwpw-194203): add data-country/data-card-url to card variants

markpadbe:
## Summary
Surface two read-only data attributes on every CaaS card's root `<li>` so Milo's MEP debug "Highlight CaaS" feature can identify regional vs. fallback content and provide a click-through target. No runtime behaviour changes for any current consumer.
- **`data-country`** — populated from the existing `country` field on the Chimera response (e.g. `"us"`, `"xx"`).
- **`data-card-url`** — populated from the existing `reference` field on the Chimera response (the canonical content URL).
Both attributes are emitted only when the corresponding value is truthy, so empty defaults don't add DOM noise.
## Why

The two new data attributes will be consumed by a separate Milo PR to power a new MEP feature, **Highlight CaaS**, shown below:

Screenshot 2026-05-07 at 2 18 34 PM

### DIFF
--- a/react/src/js/components/Consonant/Cards/__tests__/Card.spec.js
+++ b/react/src/js/components/Consonant/Cards/__tests__/Card.spec.js
@@ -143,3 +143,33 @@ describe('Card Component - Specific Conditions', () => {
         expect(descriptionTextElement).toHaveTextContent('This is a bold and italic text.');
     });
 });
+
+describe('MEP debug attributes for Hightlight CaaS feature', () => {
+    const cardStyles = [
+        'one-half', 'three-fourths', 'double-wide', 'half-height',
+        'product', 'text-card', 'full-card', 'icon-card', 'news-card',
+        'blade-card', 'editorial-card', 'blog-card', 'horizontal-card',
+        'button-card',
+    ];
+
+    test.each(cardStyles)('%s stamps data-country and data-card-url when populated', (cardStyle) => {
+        renderCard({
+            cardStyle,
+            country: 'us',
+            reference: 'https://business.adobe.com/customer-success-stories/foo.html',
+        });
+        const li = screen.getByTestId('consonant-Card');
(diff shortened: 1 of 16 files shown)


raissanjay (approved):

react/src/js/components/Consonant/Cards/__tests__/Card.spec.js
```diff
@@ -143,3 +143,33 @@ describe('Card Component - Specific Conditions', () => {
         expect(descriptionTextElement).toHaveTextContent('This is a bold and italic text.');
     });
 });
+
+describe('MEP debug attributes for Hightlight CaaS feature', () => {
```
raissanjay, line 147: Minor: typo in the describe block — "Hightlight" should be "Highlight". Worth fixing in this PR since it will appear in every test report from here on.

raissanjay (approved):

raissanjay:
Approved. Just a minor comment above. Once its addressed and CI is green, feel free to merge on your end. 

FYI this will create a github release once merged, and then every Monday we will release to Prod. So timing your merge against that window is the simplest way to get it live when you want.

react/src/js/components/Consonant/Cards/__tests__/Card.spec.js
markpadbe: @raissanjay thanks/good catch!
Local push to update is getting rejected by branch protection (the 11-status-check rule). Ok to try edit here via Github Web?

raissanjay: Yes that's fine to edit via Github web.

markpadbe:
hi @raissanjay, as discussed separately, I appear to be blocked from updating my branch to fix the typo (permissions issue), so I will ignore the typo for now, and go ahead with the merge. Thanks.  
cc: @sheridansunier

PR #465, opened 2026-05-19
feat(mwpw-186583): removes empty collection from published page

cmiqueo:
### Problem
When the CaaS API returns zero cards for a given configuration, the collection widget was still rendered on the page — leaving an empty, visually broken block visible to end users.
### Solution
Added a `removeCollectionFromPage()` helper that is called when the API responds with an empty cards payload **and** the endpoint has `originSelection=events`. The behavior is context-aware:
- **Published page** — if the component is mounted inside `div#caas.caas-preview`, the entire collection root is removed from the DOM so visitors never see the empty widget.
- **Configurator** (`div.caas-config` ancestor) — removal is explicitly skipped so authors are not left with a broken editing context.
- **No `originSelection=events`** — the function is never called, preserving existing behavior for non-events collections.
### Changes
- `Container.jsx`
  - Renamed `noResultsDescription` config read to `authoredNoResultsDescription` and promoted it to a state variable to support future runtime overrides.
  - Promoted `box` ref declaration to earlier in the component (before state declarations) so it is available during the fetch chain.
  - Added `removeCollectionFromPage()` — traverses the DOM via `closest()` to find the collection root, then removes it if not inside `.caas-config`.
  - Reads `originSelection` from the endpoint URL's search params and gates the removal call on `originSelection === 'events' && box.current`.
### Test Page
  - https://stage--milo--adobecom.aem.page/drafts/caas/qa/mwpw-186583
  - https://stage--milo--adobecom.aem.page/drafts/caas/qa/mwpw-186583?caasver=beta
### Tests
Four unit tests added to `Container.spec.js` under `describe('removeCollectionFromPage (mwpw-186583)')`:
| Test | `originSelection` | DOM wrapper | Expected |
|---|---|---|---|
| Not an events collection | `(none)` | `div#caas.caas-preview` | stays in DOM |
| Events, no recognized ancestor | `events` | none | stays in DOM |
| Events, published page | `events` | `div#caas.caas-preview` | **removed** from DOM |
| Events, inside configurator | `events` | `div.caas-config > div#caas.caas-preview` | stays in DOM |
All 35 unit tests pass.

### DIFF
--- a/react/src/js/components/Consonant/Container/Container.jsx
+++ b/react/src/js/components/Consonant/Container/Container.jsx
@@ -145,7 +145,7 @@ const Container = (props) => {
     const topPanelSearchPlaceholder = getConfig('search', 'i18n.topFilterPanel.searchPlaceholderText');
     const searchPlaceholderText = getConfig('search', 'i18n.filterInfo.searchPlaceholderText');
     const noResultsTitle = getConfig('search', 'i18n.noResultsTitle');
-    const noResultsDescription = getConfig('search', 'i18n.noResultsDescription');
+    const authoredNoResultsDescription = getConfig('search', 'i18n.noResultsDescription');
     const apiFailureTitle = getConfig('collection', 'i18n.onErrorTitle');
     const apiFailureDescription = getConfig('collection', 'i18n.onErrorDescription');
     const isLazy = getConfig('collection', 'lazyload');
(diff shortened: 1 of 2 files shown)


shkhan91 (approved):

sheridansunier (approved):
Solid, well-scoped PR with good test coverage. The three-branch context detection (published page / configurator / unknown) is clean and the tests cover each case. A few minor notes for follow-up if desired: (1) Direct DOM removal of a React-managed node — verify this doesn't cause unmount errors in production edge cases. (2) Consider `collectionRoot.remove()` over `removeChild` for modern JS consistency. (3) The `!box.current` early-return guard is currently untested. Nothing blocking. Approved.

PR #483, opened 2026-05-31
feat(mwpw-196902): developer-friendly issue body via Claude narrative rewrite

sanrai:
## What

Makes the QA Audit issue body developer-friendly instead of a raw agent transcript dump.

[Issue #482](https://github.com/adobecom/caas/issues/482) is the example of the current state: scannable but contains lines like `Turn 1/10 [claude] 1 msgs 11290b ... ok` and `> find_and_show({"selector":"..."})` that mean nothing to a CaaS engineer reviewing a defect. Most of the value is buried under noise.

## How

`aggregate-report.mjs` now does a two-stage pipeline:

1. **Parse** per-scenario logs into structured findings (verdict, turn count, agent's own final report).
2. **Rewrite via Claude** (same Adobe LLM proxy used by the agent loop) into a clean GitHub Issue body: headline, summary table, per-FAIL section with **user impact** + **what was tested** + **what was found** + **suggested fix** when obvious, brief PASS list at the end.

The raw per-scenario transcripts are preserved in a collapsed `<details>` block at the bottom for QA-tool debugging — they just aren't the primary surface anymore.

If the LLM call fails (proxy down, token expired, parse error), the script falls back to the structured-but-not-prettified format so the issue still posts.

## Verdict reliability

Verdict was previously detected by grepping the markdown table for `| FAIL |` / `| INCOMPLETE |`. Since the LLM rewrite may use different table wording, that was brittle. Replaced with a deterministic sidecar file (`/tmp/qa-audit-verdict.txt`) that aggregate-report.mjs writes BEFORE the LLM call. Workflow reads the sidecar directly. Verdict survives even if the narrative pass fails.

## Files changed

- `.github/qa/aggregate-report.mjs` — two-stage pipeline + verdict sidecar + fallback
- `.github/workflows/qa-audit.yml` — pass `IMS_ACCESS_TOKEN` to the aggregate step + read verdict from sidecar instead of grepping markdown

## Validation

Verified locally against the results of issue #482's run: the structured fallback (when LLM is unreachable) is already much more readable than what shipped. Will confirm the narrative pass quality on the next CI run after merge.


sanrai:
Added a follow-up commit (39688693) that strips the hardcoded LLM proxy URL and model name from the source. Both now read from env vars only, sourced from `secrets.LLM_PROXY_ENDPOINT` (exists) and `secrets.LLM_MODEL` (needs to be added). Also renamed `callClaude` / `rewriteWithClaude` to LLM-agnostic names so the source no longer leaks vendor choice.

**Action required before merge:** add an `LLM_MODEL` repo secret with the model name. Without it the narrative-rewrite step falls back to the structured-raw format (still readable, just not as polished).

The same model name is currently hardcoded in `ai-code-review.yml`'s Python script. Worth migrating that one too in a follow-up to consolidate.

sanrai:
Two more changes pushed (49bf280c):

1. **`LLM_MODEL` repo secret added** via gh CLI. Set to the model name we were using before. Workflow can now successfully run the narrative-rewrite step.

2. **`ai-code-review.yml` migrated to source the model from `secrets.LLM_MODEL`** too. No more hardcoded model name in either workflow. Fails fast with exit 2 if the secret is missing rather than silently posting a review using an unintended default.

Both `LLM_PROXY_ENDPOINT` (existing) and `LLM_MODEL` (just added) are now the single source of truth for LLM infra config across the repo. No vendor or model name appears in any committed file.

Ready to merge.

PR #486, opened 2026-05-31
feat(qa-audit): tiered suites + parallel scenario execution

sanrai:
## What

Adds two capabilities to the QA Audit workflow:

1. **Suite picker** — workflow now takes a `suite` input (choice of `smoke / a11y / functional / regression / all`, default `a11y`). Lets you run a fast pre-release smoke or a deep regression on demand without changing the workflow.
2. **Parallel execution** — scenarios run concurrently via `xargs -P WORKERS` (default 4). Each worker gets its own Chrome user-data-dir so they don't collide on Playwright's profile lock.

## Why

The current sequential a11y suite takes 7-13 minutes. As we add functional and regression scenarios, sequential would grow to 30-60 minutes. Parallel-4 keeps the a11y suite under ~4 minutes and a full 20-scenario regression suite under ~8 minutes — practical to wait for at a release gate.

## How it lays out

```
.github/qa/qa-prompts/
  a11y/         <- the 8 scenarios we already had (moved here)
  smoke/        <- empty placeholder, populate with 3-4 fast checks
  functional/   <- empty placeholder, populate with filter/search/sort/pagination
  regression/   <- empty placeholder, populate with combinatorial coverage
  visual-diff/  <- empty placeholder, separate workflow shape
```

When you trigger the workflow, pick a suite and the script iterates `qa-prompts/<suite>/audit-*.txt`. `all` concatenates everything across subdirs.

## Inputs added

```yaml
suite:
  type: choice
  options: [smoke, a11y, functional, regression, all]
  default: a11y
workers:
  type: string
  default: '4'  # parallel worker count
```

The existing `caasver`, `base_url`, `max_turns` inputs are unchanged.

## Implementation notes

- Each parallel worker runs `node qa-runner-v2.mjs` in its own subshell with `USER_DATA_DIR=/tmp/chrome-qa-w<pid>`. Unique per worker; subsequent workers reuse PIDs that happen to come free as previous ones exit — fine for our use case since we recreate the dir each invocation.
- `pkill -f USER_DATA_DIR` from the old sequential script is removed because per-worker dirs make global cleanup unnecessary (and dangerous in parallel — it'd kill sibling workers).
- Verdict aggregation reads each scenario's log independently; the existing `aggregate-report.mjs` works unmodified.
- Issue title and artifact name now include the suite name so multiple runs against the same caasver but different suites are distinguishable.

## Validated

Locally executed the new shell script with `SUITE=a11y WORKERS=4`. The 8 scenarios fanned out to 4 workers, all spawned + reaped cleanly, each got its own Chrome user-data-dir (no collisions). Local IMS token was expired so verdicts were NO_VERDICT, but the orchestration itself worked end to end — CI has a fresh token rotated daily.

## Follow-ups (out of scope for this PR)

- Populate `smoke/`, `functional/`, `regression/` directories with scenarios — one PR per tier.
- Build the `visual-diff` workflow as a separate file (`qa-visual-diff.yml`) since its shape is different (two URLs in, single diff scenario).
- Add a per-scenario `timeout 600` wrapper as a belt-and-suspenders defense against a single hung scenario eating the workflow budget.


sanrai:
Local validation finding: pushed a follow-up commit (0080550d) defaulting `WORKERS=1` (sequential).

**Why:** Tested parallel-2 and parallel-4 on the MacBook Air runner with a fresh IMS token. Both degrade verdict quality:

| Config | Wall time | Verdict quality |
|---|---|---|
| Sequential (workers=1) | ~720s | Accurate — true PASS / true FAIL per scenario |
| Parallel-2 | 356s ✅ | 8 of 8 scenarios FAIL — including 4 that should be PASS |
| Parallel-4 | killed at 340s | First 4 workers never finished; 56 Chrome processes saturated the Mac |

**Root cause:** the Air's CPU can't sustain multiple concurrent Playwright Chrome trees. Pages load 2-3x slower under contention, so the agent burns its 10-turn budget on observation before it can call done(). Auto-done synthesis defaults to FAIL — producing false positives that would waste engineer time investigating bugs that don't exist.

**The infrastructure is correct, the hardware isn't.** Once the runner moves to a Mac mini M4 or any beefier machine, flip `workers` back up to 4 via the workflow input. The plumbing supports it; only the default changes.

Suite picker (smoke / a11y / functional / regression / all) is unaffected by this change and stays the primary value of the PR.

Ready to merge.

sanrai:
Pushed a follow-up commit (bdeaff49) adding the visual-diff workflow + scenario.

**New workflow:** `.github/workflows/qa-visual-diff.yml` — `workflow_dispatch` with `version_a` (baseline) + `version_b` (candidate) inputs. Runs a single scenario against `qa-prompts/visual-diff/version-diff.txt`. Same self-hosted runner. Posts an Issue labeled `qa-visual-diff` + `qa-PASS`/`qa-FAIL`. ~90 seconds end to end.

**Scope philosophy baked into the prompt:** explicitly scoped to the CaaS card-collection region, with an inline IGNORE list (cookie banner, Adobe global nav, A/B-tested page chrome, sub-pixel rendering noise, mid-frame animations, dynamic timestamps, "X results" count text) and a REAL REGRESSIONS list (layout shifts, typography, color/theming, spacing, element presence, card layout proportions, broken images, pagination/filter visual changes). The agent uses the diff overlay as a focus signal, then classifies each highlighted region accordingly.

**Verdict logic:** PASS if pixel diff is essentially zero (< 0.05%) OR all differences are ignorable. FAIL with a specific regression list otherwise. Headline does not include the raw pixel count — agreed that number is meaningless to reviewers.

**Use case:** trigger before a release tag to confirm the new bundle doesn't break visible rendering vs the previous one. Or before any caasver bump on a Milo page.

This PR is now scope-complete:
- Tiered suites (workflow input)
- Parallel execution infra (default sequential per the hardware constraint)
- Visual diff workflow + scenario

Ready to review.

sanrai:
**Visual-diff scenario validated locally on two real-world version pairs.**

| Test | Pair | Expected | Got | Wall |
|---|---|---|---|---|
| 1 | 0.51.0 vs 0.51.1 (known: Sort by: label added per mwpw-177207) | FAIL flagging the sort widget | **PASS** ✅ — caught it: 0.14% diff localized to sort widget, agent identified the exact text change with severity | 84s |
| 2 | 0.51.4 vs 0.51.5 (neutral: only CI workflow changes) | PASS | **PASS** ✅ — card grid 0% diff; agent saw a 23.6% diff in the upper region and correctly identified it as page-level marketing chrome + chatbot pill (noise per IGNORE list), then scrolled deeper to verify CaaS itself was identical | 89s |

**Key validation:** test 2 had a huge raw pixel diff (23.6%) but the agent didn't naively fail on it. It applied the IGNORE list, adapted by scrolling to inspect the actual CaaS region, and correctly cleared the run. That's exactly the noise-filtering behavior the prompt was designed for.

**Two tunings learned and pushed in 801614ed:**

1. **MAX_TURNS bumped from 8 to 12.** 8 was too tight — the scenario has 7 deterministic steps (navigate A, scroll, screenshot A, navigate B, scroll, screenshot B, diff, load, analyze, done) plus the agent often inserts one or two adaptive turns when the first screenshot looks incomplete. Both validation runs landed at 10-12 turns with margin.

2. **USER_DATA_DIR switched to /tmp/chrome-qa-profile** (the warmed one already used by the audit workflow). A fresh profile failed the second navigate with `ERR_HTTP2_PROTOCOL_ERROR` because adobe.com's bot detection flagged the cookieless + rapid-second-request pattern. The warmed profile has accumulated cookies / fingerprint state that Cloudflare accepts.

Workflow is now ready. After merge, you trigger from the Actions tab with any two `caasver` values and get an Issue with either "No visual regressions detected" or a specific list of what changed.

PR #492, opened 2026-06-16
test: Mac mini self-hosted runner smoke check

raissanjay:
Temporary smoke test to verify the new ARM64 self-hosted runner on the Mac mini picks up jobs and that IMS_ACCESS_TOKEN works against the LLM proxy. Will be deleted after verification. Do not merge.

### DIFF
--- a/.github/workflows/mini-runner-smoke.yml
+++ b/.github/workflows/mini-runner-smoke.yml
@@ -1,0 +1,35 @@
+name: Mini Runner Smoke Test
+
+on:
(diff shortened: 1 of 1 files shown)


raissanjay:
Smoke test done: confirmed the ARM64 self-hosted runner on the Mac mini picks up and runs jobs. Proxy step shows IMS_ACCESS_TOKEN is currently expired (separate token-rotation work). Closing, not merging.

PR #532, opened 2026-06-22
feat(mwpw-195996): adds new icon and product name option 

cmiqueo:
## Summary
- Add a "product name" detail-text option to flex cards that swaps the eyebrow/detail text for a product icon + name when the card's detail text matches a configured product tag
- Wire the product list from collection config through `Grid` → `Card` → `FlexCard` and render it via `CardContent`
- Add `productInfoType` PropType and `.product-info` styles for the new label variant
## Jira Ticket
Resolves: [MWPW-195996](https://jira.corp.adobe.com/browse/MWPW-195996)
## Changes
- `react/src/js/components/Consonant/Grid/Grid.jsx` — reads `collection.detailsTextOption`; when set to `productName`, resolves the `products` config object into an array and passes it to each card
- `react/src/js/components/Consonant/Cards/Card.jsx` — adds `products` prop (PropType + default `[]`) and threads it into the memoized `cardData` context
- `react/src/js/components/Consonant/Cards/FlexCard.jsx` — matches the card's `detailText` against `products[].tagID`; on a match, builds `productInfo` (`tagID`, `title`, `icon` from `tagImage`) and suppresses the plain detail text
- `react/src/js/components/Consonant/Cards/CardContent/CardContent.jsx` — renders the new `product-info` label (icon + title) when `productInfo` is provided; renames the icon-alt label test id to `consonant-Card-label-icon-alt`
- `react/src/js/components/Consonant/types/card.js` — adds `productInfoType`
- `less/components/consonant/cards/flex.less` — styles for `.product-info` (icon sizing, title weight/color)
- `react/src/js/components/Consonant/Cards/__tests__/Flex.spec.js` — tests for rendering product info and hiding detail text when product info is shown
## Test plan
- [ ] With `collection.detailsTextOption = 'productName'` and a `products` map, a card whose detail text equals a product `tagID` renders the product icon + name instead of the eyebrow text
- [ ] With `detailsTextOption` unset or no matching product, the card still renders its normal detail text (regression)
- [ ] Verify the icon-alt label change (`consonant-Card-label-icon-alt`) didn't break any consumers of the old `consonant-Card-label` test id
## Usage
Under the **UI Tab** select:
 - Card Style: **Flex Card**
 - Detail Text: **Product Name**
## Screenshot
Screenshot 2026-06-22 at 2 40 27 PM


sanrai (changes requested):
AI feedback result looks valid. Please fix the following issues

1. products derived from config on every render without memoization (Grid.jsx)
2. products config key assumed to be an object/dict (Grid.jsx)
3. detailText match used as product lookup key — case-sensitive string equality (FlexCard.jsx)

react/src/js/components/Consonant/Grid/Grid.jsx
```diff
@@ -93,6 +93,11 @@ const Grid = (props) => {
     const paginationType = getConfig('pagination', 'type');
     const collectionButtonStyle = getConfig('collection', 'collectionButtonStyle');
     const cardHoverEffect = getConfig('collection', 'cardHoverEffect');
+    const detailsTextOption = getConfig('collection', 'detailsTextOption');
+    let products = [];
+    if (detailsTextOption === 'productName') {
+        products = Object.values(getConfig('products', '') || {});
```
sheridansunier: `Object.values()` creates a new array reference on every render, which will invalidate every card's `useMemo` even when nothing has changed. Consider wrapping `products` in a `useMemo` keyed on `detailsTextOption`:

```jsx
const products = useMemo(() => (
    detailsTextOption === 'productName'
        ? Object.values(getConfig('products', '') || {})
        : []
), [detailsTextOption]);
```

react/src/js/components/Consonant/Cards/CardContent/CardContent.jsx
```diff
@@ -33,9 +35,19 @@ const CardContent = ({
...
+                {productInfo.tagImage && (
+                    <img className="product-info-icon" src={productInfo.tagImage} alt={productInfo.title || ''} />
+                )}
+                <span className="product-info-title">{productInfo.title || ''}</span>
+            </span>
+        )}
         {showIconAlt && (detailText === '') &&
         <span
-            data-testid="consonant-Card-label"
+            data-testid="consonant-Card-label-icon-alt"
```
sheridansunier: This renames a previously documented test ID (`consonant-Card-label` → `consonant-Card-label-icon-alt`), which is a breaking change for any Playwright/Cypress/integration tests in consuming repos that query by that ID. Please grep for the old ID across at minimum `adobecom/caas-configurator` and `adobecom/milo` before merging — or note confirmed findings here.

sheridansunier (changes requested):
Two blocking issues inline — everything else looks good to me.

sanrai:
@ai-bot the finding "Unsafe `Object.values()` on potentially non-object config value" (`FlexCard.jsx`) is a false positive, and the shape it's flagging is **intentional**, not an oversight. The line:

```js
const products = detailsTextOption === 'productName'
    ? Object.values(getConfig('products', '')).filter((product) => product && product.tagID)
    : [];
```

was written this way on purpose as a defensive catch-all, after deliberately choosing it over the more verbose `typeof`-guard alternative. The design relies on two facts you can verify in the source:

1. **`getConfig('products', '')` never returns `null`/`undefined`.** `DEFAULT_CONFIG.products` is `{}` (see `constants.js`), and `getConfig`/`makeConfigGetter` (in `consonant.js`) falls back to the `DEFAULT_CONFIG` value when the key is unset or nullish. So a missing or null `products` config resolves to `{}`, and `Object.values({})` is `[]`. There is no throw path.

2. **The trailing `.filter((product) => product && product.tagID)` is the catch-all for malformed values.** If `products` were ever a non-object like a string, `Object.values('abc')` yields `['a','b','c']`, and the filter drops every entry lacking a `tagID`, so junk never reaches `.find()`. The filter exists precisely to neutralize the "non-object" case the finding worries about.

Case-by-case (all verifiable from the source above):

```
missing  -> getConfig returns {}       -> Object.values -> []   (safe)
null     -> getConfig returns {}       -> []                     (safe)
{ ... }  -> the products object        -> product entries        (intended)
"abc"    -> ['a','b','c'] -> filtered  -> []                     (safe, catch-all)
```

So there is no unsafe path, throw or output. This was a conscious "validate the elements, not the container" choice. Please drop this finding.

sanrai:
@cmiqueo  I went through the AI review myself and vetted each point against the actual code rather than just forwarding it.

**`getConfig` / `Object.values`.** Functionally your version worked, but it wasn't bulletproof. The robust pattern is to give `products` a `DEFAULT_CONFIG` default (`{}`) plus the element filter, so missing / null / non-object configs all resolve safely. I've made that change. So the AI's flag had a real point.

**The other two are real, I verified them, please address them.**

**1. `productName` detail logic (`Card.jsx` + `FlexCard.jsx`).** `showProductName`, `detailText`, and `productInfo` use mismatched conditions. Two concrete failures, reproduced against the real `Card -> FlexCard -> CardContent`:

```
Case 1  product w/ title   -> <img> + "Acrobat"                 (ok)
Case 2  no match           -> raw tagID renders as the label    (wrong)
Case 3  product, NO title  -> labelText:null, productHTML:null  (blank, wrong)
```

Root cause: `showProductName` flips on a match (suppressing the text), but `productInfo` only renders if it has a `title`, and `title` is optional per `productInfoType`:

```jsx
if (productData) { showProductName = true; productInfo = { tagID, title, tagImage }; }
detailText={showDetails && !showProductName ? detailText : ''}
productInfo={showDetails && productInfo && productInfo.tagID && productInfo.title ? productInfo : null}
```

Fix: align the conditions. Only set `showProductName` when a product will actually render (`title || tagImage`), and don't let a raw tagID show on no-match.

**2. `useMemo` deps (`Card.jsx`).** `extendFooterData` and `parseMarkDown` are render-body functions (new identity each render) sitting in the `cardData` dep array, so the memo recomputes every render and does nothing. I checked the rest of the array, every other dep is a primitive or stable reference, so those two are the whole problem. Wrap both in `useCallback` (deps are stable primitives; just confirm `onClick` is stable) and the memo actually works. Clean, contained fix.

You can re-run the review anytime by commenting `@ai-bot`, it'll take your input, weigh your evidence, and post a fresh review. Address these two and I'll clear my change request.

sanrai (approved):

PR #560, opened 2026-07-21
test: verify review-score gate on main (do not merge)

sanrai:
Temporary PR to confirm the gate blocks without a net approval and that checks run cleanly.


raissanjay (dismissed):
approve: verifying gate clears on main

raissanjay (approved):
re-approve: watch if the check clears

PR #563, opened 2026-07-21
demo: show review-score gate blocks (do not merge)

sanrai:
Proof PR: should be blocked by review-score-gate until it gets a net approval.


raissanjay (approved):
approve: watch the gate flip green

PR #565, opened 2026-07-23
fix(mwpw-182729): modern carousel nav controls on mobile

cmiqueo:
**Summary**
- Force the modern carousel layout on mobile viewports, even when isModernCarousel isn't set in config
- Move the mobile-detection breakpoint (TABLET_BREAKPOINT) into shared constants and compute isMobile once as a module-level export in Helpers.js instead of a per-render function in CardsCarousel
- Update CardsCarousel and Container to consume the shared isMobile value instead of local duplicate logic

**Jira Ticket**
Resolves: MWPW-182729

**Changes**
- react/src/js/components/Consonant/Helpers/constants.js — added TABLET_BREAKPOINT constant
- react/src/js/components/Consonant/Helpers/Helpers.js — added isMobile export computed from window.innerWidth < TABLET_BREAKPOINT
- react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx — removed local isMobile() function/constant, now imports and uses shared isMobile
- react/src/js/components/Consonant/Container/Container.jsx — imports isMobile, applies modern-carousel class and carouselType="modern" when on mobile regardless of isModernCarousel

**Test plan**
- [ ] Verify carousel nav controls render correctly on mobile viewports for both default and modern carousel configs
- [ ] Verify desktop/tablet carousel behavior is unchanged

**Notes**
isMobile is now computed once at module load from window.innerWidth rather than re-evaluated on each call — it won't update on viewport resize/orientation change, which differs from the previous per-invocation isMobile() function.

### DIFF
--- a/react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
+++ b/react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
@@ -83,7 +83,7 @@ function CardsCarousel({
     const prevCard = getConfig('collection', 'i18n.prevCards') || 'Previous Cards';
     const useLightText = getConfig('collection', 'useLightText');
     // eslint-disable-next-line no-use-before-define
-    const isIncremental = isMobile || getConfig('pagination', 'animationStyle').includes('incremental');
+    const isIncremental = isMobile() || getConfig('pagination', 'animationStyle').includes('incremental');
     const renderOverlay = getConfig('collection', 'useOverlayLinks');
 
     if (cardsUp.includes('2up')) {
@@ -205,7 +205,7 @@ function CardsCarousel({
     }
 
     function mobileLogic() {
-        if (isMobile && carouselType === 'default') {
+        if (isMobile() && carouselType === 'default') {
             hideNav();
         } else {
             showNav();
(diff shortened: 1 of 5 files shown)


cmiqueo:
@ai-bot the change on behavior from isMobile() to isMobile is intentional.

sanrai:
@cmiqueo are you sure this is the expected behavior? I pulled this down and noticed the mobile version only shows up if the page starts out small. If you rotate the tablet or resize the window after it loads, it doesn't switch over unless you refresh. Looks like it's because `isMobile` is set once instead of being a function. Are you sure you don't want it to stay a function? This looks like a bug to me.

image

cmiqueo:
@sanrai 
It is not a bug, but in the in the interest of time I put it back to my original function implementation.
Note that the functionality and behavior does not change from previous releases!

sanrai:
This looks good. Will approve after one small code change.

On mobile the `modern-carousel` class gets added to non-carousel layouts too (grids, paginator, etc.). I tested it and it doesn't affect the layout, but it's confusing to see it in the markup where it doesn't belong. The fix is to only add it when it's actually a carousel.

Right now it's:

```jsx
'modern-carousel': isModernCarousel || isMobile(),
```

Change it to (add `isCarouselContainer &&`):

```jsx
'modern-carousel': isModernCarousel || (isCarouselContainer && isMobile()),
```

`isCarouselContainer` is already defined right above, so it's a one line change. Make that and I'll approve. Thanks!

cmiqueo:
@sanrai done

sanrai (approved):

PR #567, opened 2026-08-05
feat(mwpw-203439): emit Schema.org JSON-LD ItemList for card collections

sanrai:
## What

Collections can now emit one `<script type="application/ld+json">` block describing the rendered cards as a Schema.org ItemList. Each entry is minimal on purpose: `url` + `keywords`.

image

## Why

Cards render as divs whose tag taxonomy is invisible to machine readers: tags arrive hashed from the backend. Titles and descriptions are already crawlable in the visible DOM, so this block adds only what is otherwise unreadable: the resolved tag classification (`keywords`) attached to a citable `url`.

## How tag resolution works

We never reverse the hash. The authored filter config pairs tag ids with human labels, and Container already hashes those ids in place to match hashed card tags (existing behavior the filter panel depends on). So the config maps (hashed) tag id -> label for free. Card tags without a filter entry are skipped, never emitted raw.

## Safety

- Opt-in: `collection.showJsonLd`, default `false`.
- Additive script tag injected in a `useEffect` after paint; scoped per collection container, so multiple collections on one page each manage their own block. Removed when the rendered card list becomes empty, so the metadata never describes cards no longer shown.
- Serializes at most 50 cards; `numberOfItems` reports the true filtered total.

## Testing

- 15 unit tests: tag map (hashed + unhashed + nested), entry building, cap and true-total, JSON round-trip, injection, idempotent re-injection, multi-collection independence, stale cleanup.
- e2e: block present and parseable with flag on, absent with flag off (run-e2e-tests green against the PR build).
- Manually verified on localhost against the hashed fixture: hashed card tags resolve to correct labels end to end.

## Known limitation

Client-rendered only: crawlers that do not execute JS will not see the block. Static-fetch coverage would require SSR, out of scope.

## Rollout

Enable on one low-traffic collection surface first. Note: keyword coverage equals the page's filter panel coverage; tags outside the panel taxonomy are skipped by design. MWPW ticket to be linked.


cmiqueo:
@sanrai 

Take a look at these findings and let me know if these are known issues so that I can approve it.

Code review(review · 7 findings)
react/src/js/components/Consonant/Container/Container.jsx
  ● 1550 [correctness] JSON-LD is built from the full filtered/timed card list (gridCards), not the subset actually rendered on the current page.
  ● 1548 [correctness] JSON-LD is injected client-side only (useEffect after mount), which defeats the feature's stated purpose for non-JS crawlers.
  ● 1549 [correctness] Toggling collection.showJsonLd from true to false at runtime leaves the stale JSON-LD block in the DOM.
  ● 1557 [efficiency]  The useEffect dependency array includes gridCards, a new array reference on every render, so it re-runs (remove+stringify+append) on every Container re-render regardless of whether cards changed.
.github/workflows/pull-request.yaml
  ●  160 [reliability] Fixed `sleep 3` before health-checking the locally served build is a race condition that can fail CI on slow runners.
  ●  158 [efficiency]  The new e2e job rebuilds the project from scratch a third time, duplicating work already done by check-build and deployment jobs in the same workflow run.
react/src/js/components/Consonant/Helpers/jsonLd.js
  ●   35 [reuse]       buildTagLabelMap re-implements category/nested-item tree-walking that already exists elsewhere in the codebase instead of reusing a shared traversal helper.

One more thing worth flagging: unit tests pass with coverage thresholds met, but Container.jsx:1550 (the injectCollectionJsonLd call site) isn't exercised by any existing Container test — not a build breaker, but worth mentioning to the PR author.

✻ Cooked for 36s

※ recap: You're reviewing PR #567 (JSON-LD for card collections); the background code review just finished with 7 findings, mainly stale/misaligned JSON-LD injection timing and content. Next: decide whether to fix these issues or comment them on the PR. (disable recaps in /config)

sheridansunier:
A few things from review, roughly in order of importance:

1. **`numberOfItems` undercounts on partial/paginated loads.** The JSON-LD effect passes `totalItems: filteredCards.length`, but `gridCardLen` is the value the code already computes as the true total during server-side partial loads (`isPartialLoad`). As written, a collection with e.g. 500 total cards but 20 fetched so far would emit `numberOfItems: 20` — contradicting the comment above it that says this reports "the true filtered total."

2. **JSON-LD `url` can diverge from the card's actual rendered link.** `buildCardEntry` in `jsonLd.js` reads `footer[0].right[0].href` unconditionally as the URL fallback, but `Card.jsx`'s own `getCtaLink` only uses that path when there's exactly one footer with exactly one right-side item — otherwise it renders no CTA at all. A multi-footer card would get a JSON-LD `url` pointing to a link that isn't actually on the page.

3. **Stale `<script data-caas-jsonld>` can be left in the DOM.** The effect returns early when `showJsonLd` is `false`, so `injectCollectionJsonLd` (the only place that removes the existing script) never runs, and there's no cleanup function either. If `showJsonLd` flips `true`→`false` at runtime without a full remount, the old structured-data script stays behind.

4. **CI: the new e2e job is still coupled to the shared deployment.** `run-e2e-tests` keeps `needs: deployment` even though it now builds and serves its own dist locally — so it's still transitively blocked by the shared/racy `github-pages` Pages environment that this change was meant to decouple from.

5. **CI: `E2E_BASE_URL` scope is wider than intended.** It's set for the whole e2e job, not just the new JSON-LD spec — so all existing specs now resolve against the locally-served dist instead of the deployed `adobecom.github.io` site. Worth confirming that's intentional, since local-serve vs. deployed can differ (caching, MIME types, base path).

6. **CI: flaky health check.** `sleep 3` before curling `localhost` has no retry/poll loop, so a slow CI runner could fail the whole gate for reasons unrelated to the PR. Also, the health check only pings a static `html/e2e/index.html`, not the actual built JS/CSS bundle, so a truncated/empty webpack output wouldn't be caught here.

7. **Minor/perf**: `filteredCards`/`gridCards` aren't memoized, so the JSON-LD effect rebuilds and re-injects the script on every re-render that recomputes them (e.g. keystrokes in search), not just when the visible card set actually changes.

#1 and #2 are the ones I'd want addressed before merge — the rest are good to track as follow-ups.

react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -1534,6 +1536,26 @@ const Container = (props) => {
...
+    /**
+     * Emits a Schema.org ItemList describing the rendered cards, so LLM
+     * crawlers and agents can classify collection content. Card tag ids
+     * (hashed or not) resolve to labels via the authored filter config,
+     * which Container has already hashed to match when isHashed is set.
+     * Additive script tag, replaced on re-render; no rendering impact.
+     * Opt-in via collection.showJsonLd; serializes at most 50 cards
+     * while numberOfItems reports the true filtered total.
+     */
+    useEffect(() => {
```
sheridansunier, line 1548: **No cleanup on unmount / toggle-off** — this `useEffect` never returns a cleanup function, and `injectCollectionJsonLd`'s removal step only runs when it's actually invoked. Since the effect body early-returns on `if (!showJsonLd) return;`, that removal path is skipped whenever `showJsonLd` goes true→false, and there's no unmount cleanup at all.

Concretely: unmounting the Container (route change, re-render without remount) or flipping `showJsonLd` off in the same mounted instance leaves the injected `<script data-caas-jsonld>` block in the DOM permanently, describing cards that are no longer shown — which directly contradicts the PR description's safety claim ("Removed when the rendered card list becomes empty").

Suggest: `useEffect(() => { if (!showJsonLd) return; const script = injectCollectionJsonLd({...}); return () => script?.remove(); }, [...]);`

react/src/js/components/Consonant/Container/Container.jsx
```diff
@@ -1534,6 +1536,26 @@ const Container = (props) => {
...
+        if (!showJsonLd) return;
+        injectCollectionJsonLd({
+            cards: gridCards,
+            filters: authoredFilters,
+            container: box.current,
+            collectionTitle: getConfig('collection', 'i18n.title'),
+            totalItems: filteredCards.length,
+        });
+    }, [gridCards, showJsonLd]);
```
sheridansunier, line 1557: **Dependency array omits variables the effect reads** — deps are `[gridCards, showJsonLd]`, but the body also reads `authoredFilters`, `getConfig('collection', 'i18n.title')`, and `filteredCards.length`.

`gridCards` is `timedCollection.length ? timedCollection : filteredCards` (line 1533) — so whenever a timed-event sort is active, `gridCards` tracks `timedCollection`, fully decoupled from `filteredCards`. In that case `filteredCards.length` (the `totalItems` passed in) can change on a re-render without `gridCards` changing, so the effect won't re-run and the injected `numberOfItems` goes stale relative to the real current filtered count.

react/src/js/components/Consonant/Helpers/jsonLd.js
```diff
@@ -0,0 +1,135 @@
...
+ * @returns {Object} - Schema.org CreativeWork
+ */
+export const buildCardEntry = (card, tagLabelMap = {}) => {
+    const entry = {
+        '@type': 'CreativeWork',
+    };
+
+    const url = getByPath(card, 'ctaLink', '')
+        || getByPath(card, 'overlayLink', '')
+        || getByPath(card, 'footer.0.right.0.href', '');
```
sheridansunier, line 62: **URL fallback misses real card link slots** — this only ever reads the `right` CTA slot (`footer.0.right.0.href`). `Cards/Card.jsx`'s actual link resolution (`getCtaLink`, lines ~296-339) also serves the card's rendered link from `center` and `alt` footer slots depending on `ctaButtonStyle`.

Any card whose real, rendered link lives in a `center` or `alt` slot (a real, supported layout) will get no `url` in its JSON-LD entry at all — silently dropping the one piece of data this PR describes as its "unique contribution" for that subset of cards. Worth reusing `getCtaLink`'s resolution (or an equivalent covering all three slots) instead of a partial ad-hoc path lookup.

.github/workflows/pull-request.yaml
```diff
@@ -153,9 +153,16 @@ jobs:
           node-version: 16.13.1
       - name: Install
         run: npm ci
+      - name: Build and serve this PR's own dist
+        run: |
+          npm run build
```
sheridansunier, line 158: **Duplicate build** — this job `needs: deployment` (which already runs `npm run build` and uploads the result as a Pages artifact a few jobs up). This step rebuilds from source instead of reusing that artifact (e.g. via `actions/download-artifact`), so every PR run now builds the same commit twice with no reuse. Not blocking, but worth a follow-up to save CI time.

.github/workflows/pull-request.yaml
```diff
@@ -153,9 +153,16 @@ jobs:
...
+          npx serve -l 5000 &
+          sleep 3
```
sheridansunier, line 160: **Fragile readiness check** — `sleep 3` before curling the backgrounded `serve` process is a fixed-timing guess rather than a retry loop. On a slower/loaded runner this could intermittently fail the E2E step for reasons unrelated to this feature. Suggest a short retry loop instead, e.g. `for i in $(seq 1 30); do curl -sf ... && break; sleep 1; done`.

sheridansunier (commented):
Reviewed via automated pass (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude, conventions angles + direct source verification). Feature is opt-in/low-blast-radius and well tested overall; flagging two correctness issues that undercut the PR's own stated goals, one staleness edge case, and two CI nits. Posting as comments for discussion, not a formal block.

sheridansunier (changes requested):
Requesting changes on the two correctness issues flagged inline (missing cleanup in the useEffect leaves a stale JSON-LD script in the DOM, and the URL fallback silently drops the url for cards using center/alt CTA slots) — see review comments above for details. The dependency-array staleness and CI items are non-blocking suggestions.

PR #569, opened 2026-08-05
fix: feature QA observation captures JSON-LD structured data blocks

sanrai:
## The bug

The Feature QA review reported FAIL on PR #567 twice, claiming no `script[data-caas-jsonld]` tag was present, while the same build provably injects it (green `run-e2e-tests` asserting the tag parses, plus manual DOM inspection on localhost).

Trace: the Step 5 observation extracts only card-level data, `.consonant-Card` nodes with `title`, `text`, and `links`. Script tags, meta tags, and container-level attributes never enter the observation. Step 6 then hands the judge only that card extraction, so any assertion about non-card DOM is structurally unverifiable and yields an automatic FAIL. The judge's own wording confirmed it: "not present in the rendered card/container data **provided**".

## The fix

The observation now also captures up to four `script[type="application/ld+json"]` blocks, each with its parent element, attribute names, and up to 1500 chars of content. They are passed to the validation judge as a separate labeled section and the PR comment reports the block count.

- Observation returns `{ cards, jsonLd }` instead of a bare card array; both downstream consumers (judge prompt, comment renderer) updated to match.
- Verified with a static consumer-consistency check (all consumers read the new shape, no stale references) and `node --check`.

## Scope note

This fixes the one blind spot that produced a false FAIL today. The general class, non-visual DOM assertions (meta tags, aria attributes, dataset attributes), still shares the limitation; a follow-up could let the planner emit an `observeSelector` the harness evaluates directly. Kept out of this PR to stay minimal.

## Expected effect

Re-running Feature QA on PR #567 after this merges should flip its verdict to PASS, since the judge will finally see the block it was asked about.


sanrai:
Folding this into #567 instead: the Feature QA workflow preserves and runs the reviewer from the PR's own merge ref, so carrying the observation fix inside #567 means the next review of #567 already uses it. Same change, cherry-picked there.

PR #573, opened 2026-08-05
feat(mwpw-0): probe D, modify existing component without tests

sanrai:
End-to-end probe for the enforcement rules in #568. This PR contains the new test-enforcement.js (branched from that PR), so its own check-test-requirements job runs the new rules against a real GitHub file listing.

EXPECT: check-test-requirements PASSES (only newly added components are gated; past code is exempt)

Throwaway PR: close without merging once the check result is recorded.


sanrai:
Probe complete, result recorded in #568.

PR #575, opened 2026-08-11
feat(mwpw-198303): removes collection from page if there are no filtered cards to display

cmiqueo:
## Description
Removes collection from page when cards load successfully but the selected Event Filter narrows results to zero

## Summary
- Removes the CaaS collection div from the page when the selected Event Filters narrow results to zero cards
- Tracks the `originSelection` query param and whether cards have finished loading via refs/state, guarding the removal to only fire for `events`-origin collections after a successful load
- Adds a regression test covering the case where cards load successfully but a subsequent filter selection narrows `filteredCards` to zero

## Jira Ticket
Resolves: https://jira.corp.adobe.com/browse/MWPW-198303

## Changes
- `react/src/js/components/Consonant/Container/Container.jsx` — adds `originSelectionRef` and `hasLoadedCards` state; new `useEffect` calls `removeCollectionFromPage()` when `originSelection` includes `events`, cards have loaded, `filteredCards` is empty, and the container (`box.current`) exists
- `react/src/js/components/Consonant/Container/__tests__/Container.spec.js` — adds test verifying the collection is removed only after cards load and a filter selection zeroes out results, and remains present while a matching card is shown

## Test plan
- [ ] Verify a normal `events`-origin collection with results renders and stays on the page
- [ ] Verify selecting an Event Filter that matches zero loaded cards removes the `#caas` container from the DOM
- [ ] Verify collections not narrowed by Event Filters (empty from the start, before load) are unaffected
- [ ] Verify `.caas-config` preview mode still skips removal

## Test Page
Using "future" servertime param:
https://www.adobe.com/events.html?servertime=1986788000000

### DIFF
--- a/react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
+++ b/react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
@@ -320,7 +320,7 @@ function CardsCarousel({
     useEffect(() => {
         mobileLogic();
         setAriaAttributes(carouselRef.current);
-    }, [cards]);
+    });
 
     return (
         <Fragment>
--- a/react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
+++ b/react/src/js/components/Consonant/CardsCarousel/CardsCarousel.jsx
@@ -320,7 +320,7 @@ function CardsCarousel({
     useEffect(() => {
         mobileLogic();
         setAriaAttributes(carouselRef.current);
-    }, []);
+    }, [cards]);
 
     return (
         <Fragment>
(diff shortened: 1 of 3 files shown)


cmiqueo:
@ai-bot 
The only time the collection is to be removed from the DOM is when the result of the "Event Filters" set in the collection configuration results on a 0 cards at load time, for example "ondemad" or upoming". Not with any other filters like products.

cmiqueo:
@ai-bot re-run
A collection should never have both an eventFilter configured and other filters (e.g. products)

sanrai (approved):

PR #577, opened 2026-08-13
ci(mwpw-204307): add build-output diff check (Layer 1, advisory QA)

sanrai:
# Add a build-output diff check (Layer 1 — advisory, non-blocking)

## What this does

Adds a CI check that answers one question for a PR: **does the shipped bundle actually change?**

It builds the base branch and the PR, normalizes away build noise, and compares the compiled output:

- **NO_CHANGE** → nothing users receive changed. This is the fast-track path for the flood of Dependabot / security / dev-dependency bumps that don't touch shipped code.
- **CHANGED** → the bundle differs. A readable unified diff + summary are uploaded as artifacts for a human (or the QA agent) to judge.

The job **always succeeds** — it never blocks a merge. It only reports.

## Why

Today the feature-QA agent skips dependency bumps because they aren't "injectable features." But a dev-dependency bump can absolutely change shipped code — e.g. `@babel/preset-react` (the exact bump in #440) is a transpiler: it decides how our JSX becomes the JavaScript that ships. "It's a devDependency" is not the same as "users see no change."

This check judges by **what actually gets built**, not by which section of `package.json` moved. If the compiled bundle is byte-identical, the change is provably behavior-neutral for what ships — a real proof, not a guess. If it isn't, we get the diff to look at.

## How it works (the two layers)

**Layer 1 (this PR) — deterministic, no AI.** A plain script builds both sides, hashes the output, and produces a diff when it differs. Cheap, cacheable, no tokens. Its whole output is one status (`NO_CHANGE` / `CHANGED`) plus, when relevant, the diff artifacts.

**Layer 2 (next PR) — the QA agent.** It *consumes* Layer 1's result. On `NO_CHANGE` it fast-tracks; on `CHANGED` it reads the diff artifact **and** screenshots the page, then judges. Two pieces of evidence: the diff catches invisible logic changes, the screenshot catches visible breakage. (A screenshot alone isn't enough — a changed event handler or dropped analytics call can look identical on screen.)

Keeping them separate means the heavy build/diff stays out of the agent, and the `NO_CHANGE` signal is reusable by branch-protection rules or humans without invoking the agent.

## Files

- `.github/workflows/output-diff-check.yml` — the CI job (build base, build PR, compare, publish summary + artifacts).
- `scripts/compare-build-output.sh` — the deterministic compare (normalize → hash → diff). Runnable locally.

## Two details worth knowing

1. **Banner normalization.** `webpack.config.js` injects a `Chimera UI Libraries - Build <version> (<timestamp>)` banner on every build. That timestamp changes every run, so a naive diff would flag *every* PR as changed. The script strips that banner line before comparing. (If other non-deterministic lines ever appear, add them to `normalize()`.)

2. **Dependabot token limits.** GitHub gives Dependabot-triggered `pull_request` workflows a read-only token and withholds secrets, so Phase 1 deliberately does **not** post a PR comment. Results go to the **Actions run summary** and **artifacts**, which work for Dependabot PRs. Posting into the QA agent's PR comment is Phase 2, via a `workflow_run`-triggered job that runs with write permissions in the base-repo context.

## Scope / how to widen

Phase 1 is scoped via `paths:` to `package.json` / `package-lock.json` so it targets the dependency/security firehose first and keeps double-build cost down during the trial. **Delete the `paths:` block** to run it on every PR — the check is generic; any PR whose bundle is unchanged earns the same fast-track.

## How to test it

Locally:

```bash
# build the base branch, save its dist
git checkout main && npm ci && npm run build
mkdir -p /tmp/base && cp dist/main.js dist/app.css /tmp/base/

# build this branch, save its dist
git checkout <this-branch> && npm ci && npm run build
mkdir -p /tmp/head && cp dist/main.js dist/app.css /tmp/head/

# compare
scripts/compare-build-output.sh /tmp/base /tmp/head /tmp/out
cat /tmp/out/summary.md
```

In CI: open it against a few recent Dependabot PRs (including the `@babel/preset-react` #440 bump) and check the run summary. Expect either a clean `NO_CHANGE` or a `CHANGED` with a sane, readable diff.

## Not in this PR (on purpose)

- No PR comment yet (Phase 2 / workflow_run).
- No auto-merge wiring yet (Phase 3 — gate auto-merge on `dependabot` + `NO_CHANGE`).
- No screenshot/visual step yet (that's Layer 2, the agent).

### DIFF
--- a/.github/workflows/output-diff-check.yml
+++ b/.github/workflows/output-diff-check.yml
@@ -1,0 +1,98 @@
+# Build Output Diff — Layer 1 (deterministic, advisory, non-blocking)
+#
+# Answers one question for a PR: does the shipped bundle actually change?
+# It builds the base branch and the PR, normalizes away the build-timestamp
+# banner, and compares the output.
+#
+#   NO_CHANGE -> nothing users receive changed. Safe to fast-track. This is the
+#                auto-greenlight path for the flood of Dependabot / security /
+#                dev-dependency bumps that don't touch shipped code.
+#   CHANGED   -> the bundle differs. The unified diff + summary are uploaded as
+#                artifacts for a human (or Layer 2, the QA agent) to judge.
+#
+# This job ALWAYS succeeds — it never blocks a merge. It only reports.
+#
+# Phase 1 scope: results go to the Actions run summary + artifacts, which works
+# even for Dependabot PRs (GitHub gives those a read-only token and withholds
+# secrets, so we deliberately do NOT try to post a PR comment here). Wiring the
+# result into the QA agent's PR comment is Phase 2 (via workflow_run, which runs
+# with write permissions in the base repo context).
+
+name: Build Output Diff (advisory)
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    # Phase 1: start with the Dependabot / security firehose. Delete this
+    # `paths:` block to make the check run on EVERY PR (it's generic — any PR
+    # whose bundle is unchanged earns the same fast-track).
+    paths:
+      - 'package.json'
+      - 'package-lock.json'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: output-diff-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  build-output-diff:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0   # need base branch history to build the base ref
+
+      - name: Record PR head SHA
+        run: echo "HEAD_SHA=$(git rev-parse HEAD)" >> "$GITHUB_ENV"
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 16.13.1
+          cache: 'npm'
+
+      # ---------- Build the PR head ----------
+      - name: Install deps (PR)
+        run: npm ci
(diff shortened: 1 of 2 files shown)


cmiqueo (approved):

sheridansunier (approved):
A few concerns from review — none blocking, but worth tracking as follow-ups:

1. **The diff check compares the wrong artifact.** `compare-build-output.sh` hashes `dist/main.js`, but `index.html` actually loads `dist/main.min.js`, which is produced by a separate `postbuild` minify step that's never compared. A change that only affects the minifier (e.g. its own version bump) would report `NO_CHANGE` even though the bundle users actually receive changed — the same class of false negative the `@babel/preset-react` example in the PR description is meant to catch, just one step downstream.

2. **Build/install failures can fail the whole job.** There's no `continue-on-error` on the install/build steps in `output-diff-check.yml`, which contradicts the PR's stated guarantee that "the job always succeeds — it never blocks a merge." A Dependabot bump that breaks `npm ci`/`npm run build` (exactly the PR type this check targets) will turn the job red instead of producing an advisory report.

3. **Silent failure can mask a broken build as `NO_CHANGE`.** The `cp -f ... 2>/dev/null || true` swallows copy failures, and `normalize()` treats a missing file as empty. If both base and PR builds fail to produce output, both sides hash-match on empty content and report `NO_CHANGE` — silently hiding a build breakage rather than surfacing it.

None of these block this PR for me — happy to approve and track these as fast-follows if that works.

PR #579, opened 2026-08-15
ci: run build-output-diff on all PRs (drop paths filter)

sanrai:
Removes the paths filter so build-output-diff runs on every PR, not just dependency bumps. Every PR now gets a NO_CHANGE / CHANGED verdict (comparing both main.js and app.css). This closes the gap where non-dependency PRs (CI/tooling, refactors) were reported as SKIPPED because no verdict was available. deps-dev auto-pass is unchanged; human PRs still require review. Note: check-pr-title will fail without an mwpw JIRA scope in the title.

### DIFF
--- a/.github/workflows/output-diff-check.yml
+++ b/.github/workflows/output-diff-check.yml
@@ -12,12 +12,12 @@
 
 name: Build Output Diff (advisory)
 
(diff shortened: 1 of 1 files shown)


sanrai:
Folded into #578; closing this separate PR.

PR #583, opened 2026-08-19
ci(mwpw-204835): add safe Dependabot auto-merge

sanrai:
## Summary

- run root npm Dependabot checks daily at 6:00 AM Pacific with explicit automatic rebasing
- add an observation-first controller for byte-identical, Agent-QA-approved dependency PRs
- recover stale conflicts with bounded rebase/recreate requests and human escalation
- record raised-to-decision timing on each evaluated Dependabot PR

## Initial safety policy

Auto-merge eligibility is intentionally narrow:

- Dependabot-authored commits only
- direct development dependency patch/minor updates only
- only `package.json` and `package-lock.json` changed
- branch contains latest `main`
- `build-output-diff` reports `NO_CHANGE`
- Agent QA reports `PASS` for the current head SHA
- deterministic build, lint, unit, coverage, E2E, accessibility, and performance checks all pass

Anything else stays open with `dependencies-needs-human`. Only one eligible PR is armed at a time. Every rebase produces a new SHA that must pass the complete policy again.

## Rollout

The workflow defaults to observation mode and will not merge. After reviewing several decisions:

1. Enable **Allow auto-merge** in repository settings.
2. Make required checks strict/up-to-date in the `review-gate` ruleset.
3. Set repository variable `DEPENDABOT_AUTOMERGE_MODE=merge`.

The controller refuses merge mode if either repository safety setting is absent.

## Conflict recovery

- wait 30 minutes for Dependabot automatic rebase
- request `@dependabot rebase`
- after two hours, request `@dependabot recreate` once for untouched bot branches
- after another two hours, label for human intervention

Branches with human commits are never recreated automatically.

## Validation

- `node --test .github/dependabot/safe-automerge.test.mjs`
- JavaScript syntax checks for the controller and Agent QA reviewer
- YAML parse validation
- `git diff --check`
- live-data format verification against recent Dependabot PR #580

`MWPW-204307` is reused because this extends that ticket’s byte-identical build-output safety layer; retitle if this rollout has a new ticket.

### DIFF
--- a/.github/dependabot/safe-automerge.mjs
+++ b/.github/dependabot/safe-automerge.mjs
@@ -1,0 +1,355 @@
+#!/usr/bin/env node
+/**
+ * Conservative Dependabot controller.
+ *
+ * Only pure dev-dependency minor/patch PRs are eligible. They must touch only
+ * the root npm manifest/lockfile, include latest main, produce a byte-identical
+ * shipped bundle, receive a current Agent QA PASS, and pass the full PR suite.
+ *
+ * Runtime I/O intentionally goes through `gh`; pure policy helpers are exported
+ * for unit tests. This file is always executed from trusted default-branch code
+ * by workflow_run/schedule, never from the Dependabot branch.
+ */
+import { execFileSync, spawnSync } from 'node:child_process';
+import { pathToFileURL } from 'node:url';
+
+const DEPENDABOT = 'dependabot[bot]';
+const ALLOWED_FILES = new Set(['package.json', 'package-lock.json']);
+const REQUIRED_CHECKS = [
+  'Adobe CLA Signed?',
+  'agent-review',
+  'check-build',
+  'check-coverage-thresholds',
+  'check-linting',
+  'check-test-requirements',
+  'deployment',
+  'run-accessibility-checks',
+  'run-core-web-vitals-checks',
+  'run-e2e-tests',
+  'run-unit-tests',
+];
+const REQUIRED_STATUSES = ['review-score-gate'];
+const DECISION_MARKER = '<!-- dependabot-safe-automerge -->';
+const CONFLICT_MARKER = '<!-- dependabot-conflict-recovery:';
+const TWO_HOURS_MS = 2 * 60 * 60 * 1000;
+const THIRTY_MINUTES_MS = 30 * 60 * 1000;
+
+export function extractAgentVerdict(comments, headSha) {
+  const shortSha = headSha.slice(0, 7);
+  const comment = comments.find(({ body = '' }) => body.includes('<!-- agent-qa-review -->'));
+  if (!comment) return { verdict: 'MISSING', current: false };
+  const match = comment.body.match(/<!-- qa-verdict-b64:\s*([A-Za-z0-9+/=]+)\s*-->/);
+  if (!match) return { verdict: 'MISSING', current: false };
+  try {
+    const state = JSON.parse(Buffer.from(match[1], 'base64').toString('utf8'));
+    // New comments carry headSha explicitly. The header fallback keeps the
+    // first rollout compatible with comments created before this change.
+    const reviewedSha = state.headSha || ((comment.body.match(/commit\s+`([0-9a-f]{7,40})`/i) || [])[1] || '');
+    return { verdict: state.verdict || 'UNKNOWN', current: reviewedSha.startsWith(shortSha) };
+  } catch {
+    return { verdict: 'UNKNOWN', current: false };
+  }
+}
+
+export function extractDependencyMetadata(commits) {
+  const dependencyTypes = [];
+  const updateTypes = [];
+  for (const commit of commits) {
+    const body = commit.messageBody || '';
+    for (const match of body.matchAll(/dependency-type:\s*"?([^\s"\n]+)"?/g)) dependencyTypes.push(match[1]);
+    for (const match of body.matchAll(/update-type:\s*"?([^\s"\n]+)"?/g)) updateTypes.push(match[1]);
+  }
+  return { dependencyTypes, updateTypes };
+}
+
+export function hasOnlyDependabotCommits(commits) {
+  return commits.length > 0 && commits.every((commit) =>
+    (commit.authors || []).length > 0 && commit.authors.every((author) => author.login === DEPENDABOT));
+}
+
+function resultState(collection, name) {
+  const item = collection.find((entry) => (entry.name || entry.context) === name);
+  return item ? String(item.conclusion || item.state || item.status || '').toUpperCase() : 'MISSING';
+}
+
+export function evaluateCandidate(candidate) {
+  const {
+    files, commits, comments, headSha, behindBy, mergeStateStatus,
+    checkRuns, statuses,
+  } = candidate;
+
+  if (!files.length || files.some((file) => !ALLOWED_FILES.has(file.path))) {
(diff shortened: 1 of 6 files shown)


.github/dependabot/safe-automerge.mjs
```diff
@@ -0,0 +1,355 @@
...
+  if (!prs.length) {
+    console.log('No open Dependabot PRs.');
+    return;
+  }
+
+  const baseSha = gh(['api', `repos/${repo}/git/ref/heads/${defaultBranch}`, '--jq', '.object.sha']);
+  const candidates = [];
+  for (const pr of prs.sort((a, b) => Date.parse(a.createdAt) - Date.parse(b.createdAt))) {
+    if (pr.baseRefName !== defaultBranch) continue;
+    const view = ghJson(['pr', 'view', String(pr.number), '--repo', repo, '--json', 'files,commits,comments']);
```
sheridansunier: A single failed GitHub API call for one candidate PR (view/check-runs/status/compare) throws and aborts the whole evaluation run, leaving every later PR in the loop unlabeled and unevaluated.

`gh()` throws on any non-zero exit and none of the four `ghJson()` calls here pass `allowFailure:true`. If one PR hits a transient GitHub 502/rate-limit, or was closed/deleted between `pr list` and this per-PR fetch, the uncaught exception propagates out of the `for` loop in `main()` — so a later PR in the batch, possibly fully eligible, never gets evaluated, labeled, or armed for that tick. It silently waits until the next scheduled run 30 minutes later, with no per-PR isolation.

.github/dependabot/safe-automerge.mjs
```diff
@@ -0,0 +1,355 @@
...
+  }
+  if (!/NO_CHANGE/i.test(buildDiff.description || '')) {
+    return { state: 'human', reason: `shipped build output is not byte-identical (${buildDiff.description || 'unknown'})` };
+  }
+
+  const agent = extractAgentVerdict(comments, headSha);
+  if (!agent.current) return { state: 'waiting', reason: 'waiting for Agent QA on the current commit' };
+  if (agent.verdict !== 'PASS') return { state: 'human', reason: `Agent QA verdict is ${agent.verdict}` };
+
+  for (const name of REQUIRED_CHECKS) {
```
sheridansunier: `resultState()`'s FAILURE list omits the terminal check-run conclusions `skipped` and `neutral`, so a required check that legitimately completes in either state leaves the PR stuck reporting `waiting` forever instead of routing to human review.

GitHub check-runs can conclude `skipped` or `neutral` as a final, non-retriable state (e.g. a path-filtered or conditionally-skipped job). Since these aren't in the FAILURE set and also aren't `SUCCESS`, the loop returns `{state:'waiting'}` for that required check every run, and the PR never reaches `human` or `eligible` — silently stranded. Worth double-checking since the required jobs here (run-e2e-tests, run-accessibility-checks, run-core-web-vitals-checks, etc.) are plausible candidates for a future lockfile-only skip condition given this policy exists specifically for lockfile-only PRs.

.github/dependabot/safe-automerge.mjs
```diff
@@ -0,0 +1,355 @@
...
+    if (current.has(label)) gh(['pr', 'edit', String(pr), '--repo', repo, '--remove-label', label], { allowFailure: true });
+  }
+  if (add && !current.has(add)) gh(['pr', 'edit', String(pr), '--repo', repo, '--add-label', add]);
+}
+
+function disableAutoMerge(repo, pr) {
+  gh(['pr', 'merge', String(pr), '--repo', repo, '--disable-auto'], { allowFailure: true });
+}
+
+function postOnce(repo, pr, marker, body) {
```
sheridansunier: Decision-comment dedup is keyed on marker text containing only `(state, headSha)`, so if the blocking reason changes while state stays `human` for the same head, the stale original reason is never updated on the PR.

Example: head SHA X first fails with 'check-linting concluded FAILURE' → a comment is posted with marker `state:human;head:X`. Linting gets fixed but Agent QA then fails for a different reason on the same commit — overall state is still `human`, so `postOnce` finds the existing marker and skips posting. The PR permanently shows the original, now-inaccurate blocking reason even though the actual current blocker has changed.

.github/dependabot/safe-automerge.mjs
```diff
@@ -0,0 +1,355 @@
...
+  const prs = ghJson(['pr', 'list', '--repo', repo, '--state', 'open', '--author', 'app/dependabot', '--limit', '100',
+    '--json', 'number,title,baseRefName,headRefOid,mergeStateStatus,createdAt,updatedAt,autoMergeRequest,url,labels']) || [];
+  if (!prs.length) {
+    console.log('No open Dependabot PRs.');
+    return;
+  }
+
+  const baseSha = gh(['api', `repos/${repo}/git/ref/heads/${defaultBranch}`, '--jq', '.object.sha']);
+  const candidates = [];
+  for (const pr of prs.sort((a, b) => Date.parse(a.createdAt) - Date.parse(b.createdAt))) {
```
sheridansunier: Minor/efficiency: all GitHub API lookups (view, check-runs, status, compare) use blocking spawnSync/execFileSync issued strictly sequentially per PR, even though the four calls for a given PR are independent of each other and of other PRs.

With up to 5 open Dependabot PRs (per `open-pull-requests-limit`) × 4 sequential round-trips each, every scheduled tick pays full serialized network latency instead of running independent lookups concurrently — stretching wall-clock time within the 10-minute job timeout for no correctness benefit.

sheridansunier (commented):
Reviewed via automated code review. 4 findings below, most-severe first — the first two (silent run-abort and stuck-forever states) are worth fixing before merge; the last two are lower severity.

.github/dependabot/safe-automerge.mjs
sanrai: Fixed in 73cd7289. Each PR is now isolated in its own error boundary. Its four independent reads run in parallel, later PRs are still evaluated and labeled, and any incomplete read makes the run fail closed before a new auto-merge can be armed.

.github/dependabot/safe-automerge.mjs
sanrai: Fixed in 73cd7289. Decision comments are now updated when the reason changes for the same state and head SHA, while an unchanged comment is left alone. A focused regression test covers both cases.

.github/dependabot/safe-automerge.mjs
sanrai: Fixed in 73cd7289. Skipped and neutral required checks now route to human review instead of waiting forever. Startup failure is handled the same way, and the skipped/neutral behavior has a focused regression test.

.github/dependabot/safe-automerge.mjs
sanrai: Fixed in 73cd7289. The four independent reads for each PR now run concurrently with Promise.all and async gh calls. PRs themselves remain sequential so the controller keeps API pressure bounded.

sanrai:
Update: fixed in `96f5250`.

- A failed `gh pr merge --disable-auto` command now stops the controller instead of being ignored.
- The in-memory auto-merge state is cleared only after GitHub confirms the disable succeeded.
- Regression tests cover both failure and success paths.
- AI Code Review now reports 0 open findings.

This remains safe in observation mode and is now hardened before merge mode can be enabled.

sheridansunier (approved):

PR #589, opened 2026-08-20
feat(mwpw-204921): adds support for Product Icon in detailsText option

cmiqueo:
**Summary**
Add support for a new productIcon detail-text option alongside existing productName, showing only the product tag icon without the title
Card.jsx / FlexCard.jsx now match any detailsTextOption starting with product instead of the exact productName string
CardContent accepts a new productIconOnly prop to suppress rendering the product title span when only the icon should show

**Jira Ticket**
Resolves: [MWPW-204921](https://jira.corp.adobe.com/browse/MWPW-204921)

**Changes**
`react/src/js/components/Consonant/Cards/Card.jsx `— 
 broaden detailsTextOption check from `=== 'productName'` to `.startsWith('product')`
`react/src/js/components/Consonant/Cards/CardContent/CardContent.jsx` — 
 add `productIconOnly` prop; skip rendering product-info-title span when set
`react/src/js/components/Consonant/Cards/FlexCard.jsx` —
 compute `showProductIconOnly` from `detailsTextOption === 'productIcon'`, rename `showProductName` → `showProductInfo`, pass productIconOnly through to CardContent

**Test plan**
 [  ] Set detailsTextOption: 'productName' and confirm product icon + title still render as before
 [  ]  Set detailsTextOption: 'productIcon' and confirm only the icon renders (no title span)
 [  ]  Confirm non-product detailsTextOption values (e.g. date, hidden) are unaffected
 [  ]  Unit tests added/updated

**Notes**

### DIFF
--- a/react/src/js/components/Consonant/Cards/Card.jsx
+++ b/react/src/js/components/Consonant/Cards/Card.jsx
@@ -247,7 +247,7 @@ const Card = (props) => {
         detailText = staticDate.toLocaleDateString();
     } else if (detailsTextOption === 'hidden') {
         detailText = '';
-    } else if (detailsTextOption === 'productName' && cardStyle !== 'flex-card') {
+    } else if (detailsTextOption?.startsWith('product') && cardStyle !== 'flex-card') {
         detailText = '';
     }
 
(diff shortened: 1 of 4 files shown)


sheridansunier (approved):
LGTM — the `.startsWith('product')` generalization is safe (detailsTextOption is always defaulted to a string) and well-covered by the Card/CardContent/FlexCard changes.

Non-blocking note: the merged commit message is missing the `mwpw-` Jira scope your CLAUDE.md conventions call for (`feat(204921)` vs `feat(mwpw-204921)`) — worth keeping an eye on for future commits so Jira-linking/release-tagging automation doesn't miss it.

